PR #37, opened 2023-02-25
fix|build(deps): Fix requirements building

cragwolfe:
* Remove `email` from base.in, it's already in Python. otherwise get confusing pip/install errors!
* Fixes to `make pip-compile`:
  * ensures all dependencies between base and test are consistent
  * always attempts to upgrade packages with --upgrade
* Updates make test install target to just use `test.txt` (since it is a superset of `base.txt`)
* most recent version of Unstructured is now installed (now at ==0.4.15)


ajjimeno:
I have seen a few things. Probably not all due to the current PR, but just wondering how we should go about it.

1)
Problem: when running make install I got:

make: *** No rule to make target `install-install-test', needed by `install'.  Stop.

I could fix it locally changing to install-text.

2)

Problem: when running make install

ERROR: Could not find a version that satisfies the requirement nvidia-cublas-cu11==11.10.3.66 (from versions: 0.0.1.dev5)
ERROR: No matching distribution found for nvidia-cublas-cu11==11.10.3.66

Solution: I run make pip-compile

3)
Problem: when I run make pip-compile after creating a new environment following the instructions I got:

pip-compile --upgrade requirements/base.in
pyenv: pip-compile: command not found

The `pip-compile' command exists in these Python versions:
  3.8.14/envs/invoices
  3.8.14/envs/paddleocr
  3.8.14/envs/unstructured
  3.8.14/envs/utest
  invoices
  paddleocr
  unstructured
  utest

Note: See 'pyenv help global' for tips on allowing both
      python2 and python3 to be found.
make: *** [pip-compile] Error 127

Solution:

Run pip install pip-tools

4) When running make run-web-app I got: 

ImportError: failed to find libmagic.  Check your installation

Solution: 

pip install libmagic

5) We could have the example below instead of the one in the README.md since it points to the file in sample-docs

curl -X 'POST' \
  'http://localhost:8000/general/v0.0.4/general' \
  -F 'files=@sample-docs/family-day.eml' \
  | jq -C . | less -R

cragwolfe:
> pip-compile --upgrade requirements/base.in
> pyenv: pip-compile: command not found

yeah, this is an annoyance. we really should have a separate target for `pip-compile` -- this is an issue that exists in other repos too. it doesn't belong in project deps because `pip-tools` 1/ it is not needed for running the app or any tests and 2/ it pulls in `pip` which really doesn't belong in `requirements/*.txt`. so the work around until properly fixed is `pip install pip-tools`.

ajjimeno (approved):
LGTM

PR #39, opened 2023-02-27
docs: Dockerfile updated

LaverdeS:
-  Added `gcc`, `poppler`, `libreoffice`, and `detectron2` `RUN yum` commands for installing on image.
The image is generate with WARNING when run on MAC M1:
`The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested`


LaverdeS:
The image builds and runs successfully, and is properly parsing the documents in `sample-docs`

LaverdeS:
The docker image is successfully parsing `.html`, `.eml`, and `.pptx` files. For the other files in `sample-docs`:
- `.xlsx`: ValueError: Invalid file. File type not support in `partition.auto`
- `.txt`: TypeError: cannot use a string pattern on a bytes-like object in `split_by_paragraph` in `partition.text`
- `.docx`: AttributeError: 'SpooledTemporaryFile' object has no attribute 'seekable'

MthwRobinson (commented):
Couple comments so far:

I'm getting this error when I run a `.docx` file through:

```
curl -X 'POST' \
  'http://localhost:8000/general/v0.0.4/general' \
  -F 'files=@fake.docx' \
  | jq -C . | less -R
```

```python
AttributeError: 'SpooledTemporaryFile' object has no attribute 'seekable'
```

and this error with `.ppt` and `.doc`. This one can be resolved by bumping the `unstructured` version. You'll also need `libreoffice` as another system dependency for this one.

```
curl -X 'POST' \
  'http://localhost:8000/general/v0.0.4/general' \
  -F 'files=@fake.doc' \
  | jq -C . | less -R
```

```python
ValueError: Invalid file. File type not support in partition.
```

When I run a PDF through the first time, I get this. We can probably cache that model in the container so we don't need to download during the first API call.

```
Downloading (…)"model_final.pth";:  38%|████████████▏                   | 126M/330M [00:13<00:19, 10.3MB/s]
```

cragwolfe:
Related to downloading models, see: https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/ingest/doc_processor/generalized.py#L12

I.e., I think you basically need to run that python in the Dockerfile.

That still won't stop it from hitting the huggingface website though to check a config file or something when the process starts though 😿

cragwolfe:
Note on my commits:
* Notebook fixed to pass CI
* Tests added for sample-docs (these tests should also be run in CI against the built Dockerfile). 4 out of 15 are failing:
```
================================================================================= short test summary info =================================================================================
FAILED test_general/api/test_app.py::test_general_api[fake.doc] - ValueError: Invalid file. File type FileType.UNK not supported in partition.
FAILED test_general/api/test_app.py::test_general_api[fake-power-point.ppt] - ValueError: Invalid file. File type FileType.UNK not supported in partition.
FAILED test_general/api/test_app.py::test_general_api[fake-text.txt] - TypeError: cannot use a string pattern on a bytes-like object
FAILED test_general/api/test_app.py::test_general_api[fake-excel.xlsx] - ValueError: Invalid file. File type FileType.XLSX not supported in partition.
======================================================================== 4 failed, 12 passed, 2 warnings in 56.70s ========================================================================

```

cragwolfe (approved):

awalker4:
Per CORE-751, going to close this and pull the stuff we want into a new branch.

awalker4:
After walking through the diff it looks like the only thing to actually merge in here is a new test. Plus, pipeline_api accepting `file_content_type`. @cragwolfe is that something we want?

PR #45, opened 2023-03-06
Fix: pass in filename, all doc types working

LaverdeS:
Some updates to the API up to the squashed commits and [changes](https://github.com/Unstructured-IO/pipeline-general/commit/c5278957ae2800ccf4d41426b0d5e696d47a6510) by @MthwRobinson in this branch:
- `pipeline-general.ipynb` has the updates plus some fixes:
   - In `pipeline_api(...)` -> `_filename = os.path.join(tmpdir, filename)`: In case the latter `filename` starts with a slash when used in a notebook as shown in `pipeline-general.ipynb`, it will be considered an "absolute path" and everything before them is discarded. The method now handles that and works only with the filename (not path) (before `filename = /Users/username/docs/morepaths/.../fake-email.eml`; now `filename = fake-email.eml`).
   - Further cells make use of `filaname` param (necessary when on python-code) and to open as `file` in `rb`.

On the web-app run by `make run-web-app`, all document types in `samples-docs` are being parsed successfully! 🎊


LaverdeS:
Inside the docker container everything works fine besides parsing `.ppt` that raises `Internal Server Error`. Full stacktrace is inside [this gist:](https://gist.github.com/LaverdeS/567796e9dd2c19c0faeddb95af02c923)

How to reproduce:
- run `make docker-build` inside env on terminal.
- run `make docker-start-api`
- call the api:
```
curl -X 'POST' 'http://localhost:8000/general/v0.0.4/general' -H 'accept: application/json' -H 'Content-Type: multipart/form-data' -F 'files=@sample-docs/fake-power-point.ppt'
```

For `.pdf/.png|.jpg` types, It should work fine (I cannot test due to Mac-Linux tesseract Hardware Incompatibility issues). Unstructured-containerised documentation will be clear about correct use of the image/web-app.

awalker4:
Added a fix for the notebook consistency issue. Can someone give a tldr of `make api-check` and how I can account for the notebook change?

cragwolfe:
Need to run:

```
make generate-api
```
to pass the CI check (and regenerate the API from the notebook)

awalker4:
Awesome, we're green! Building the container now to investigate the `.ppt` issue above. Is this a good state to merge and go from there?

Also, earlier we mentioned getting the best of both prs - are there pieces of either one that shouldn't be included or was that meant as we just need to get both merged?

cragwolfe (approved):

cragwolfe:
CI is green, so, quick! merge it! 😆 

Yeah, unitttests from other PR can be added separately. Not sure there notebook changes there should be merged in.

Issue #46, opened 2023-03-06
Add test script(s) for global coverage > 95%


LaverdeS:
Following the tests done for `partition` in Unstructured main repo [here](https://github.com/Unstructured-IO/unstructured/tree/main/test_unstructured/partition), the idea is to create some simple tests that make use of the documents in `sample-docs` to execute most of the capabilities of the api:

Definition of done:
- Test coverage > 95% target `make test`(update if necessary).
- Include all document types in the tests even though coverage goal has been achieved.

PR #48, opened 2023-03-09
test: Add additional tests from #39

awalker4:
Closes CORE-751. Most of the stuff in the previous pr has already made it to main, just some tests left over.

Verified that @MthwRobinson's curl commands in #39 are working as well.

Edit - tests seem to hang again - will poke around a bit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ run-web-app:
 ## test:                        runs core tests
 .PHONY: test
 test:
-	PYTHONPATH=. pytest test_${PIPELINE_PACKAGE} --cov=${PACKAGE_NAME} --cov-report term-missing
+	PYTHONPATH=. pytest -vv test_${PIPELINE_PACKAGE} --cov=${PACKAGE_NAME} --cov-report term-missing
 
 .PHONY: check-coverage
 check-coverage:
--- a/Makefile
+++ b/Makefile
@@ -103,11 +103,12 @@ run-web-app:
 ## test:                        runs core tests
 .PHONY: test
 test:
-	PYTHONPATH=. pytest -vv test_${PIPELINE_PACKAGE} --cov=${PACKAGE_NAME} --cov-report term-missing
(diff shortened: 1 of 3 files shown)


cragwolfe:
> Edit - tests seem to hang again - will poke around a bit.

You may have better luck repro'ing on linux. Interestingly, also saw hanging behavior for something else -- potentially related to asyncio within multiprocessing: https://github.com/Unstructured-IO/unstructured/pull/318#issuecomment-1460769759

awalker4:
tenor

I guess that was a pretty decent onboarding adventure! Looks like the CI machine was missing some OS dependencies. I couldn't say why the result was a hang - tried this out on ubuntu and removing tesseract, etc, throws the correct error.

cragwolfe:
> Looks like the CI machine was missing some OS dependencies. I couldn't say why the result was a hang - tried this out on ubuntu and removing tesseract, etc, throws the correct error.

That's pretty wild. Nice debugging, though!

Makefile
```diff
@@ -105,9 +105,10 @@ run-web-app:
 test:
 	PYTHONPATH=. pytest test_${PIPELINE_PACKAGE} --cov=${PACKAGE_NAME} --cov-report term-missing
 
+# Setting a low bar here - need more tests!
 .PHONY: check-coverage
 check-coverage:
-	coverage report --fail-under=90
+	coverage report --fail-under=60
```
cragwolfe, line 111: this is good for now, but also worth creating an issue to increase the coverage.

i think that's pretty easily achievable by posting multiple files at once in a single request and also testing some non-200 status code cases.

cragwolfe (approved):
LGTM!

For posterity, the other PR's notebook had a way of defining a `starlette` `UploadFile` which is going to be a more realistic File-like object to test against since that is basically what is coming in from the FastAPI route when it calls `pipeline_api`. However, I think we can avoid including that complexity in the notebook for now.

image

also saved [here](https://gist.github.com/cragwolfe/4ef326139fd6b90a9653e865c57e7161).

Makefile
awalker4: Makes sense! Looks like we can use #46 to track this.

PR #49, opened 2023-03-14
chore: add initial smoke test running against containerized api

awalker4:
Closes CORE-735. Update the previously checked in dockerfile and verify in ci with a new `make docker-test` target. Add a bare bones smoke test that's basically a copy of the bare bones unit test. Without docker build caching, this takes around 15m to run (I expected worse).


awalker4:
For future reference, I found [this page](https://docs.docker.com/build/cache/backends/gha/) on caching docker builds in gh actions. That will definitely be a low hanging fruit for improving the CI runtime.

scripts/smoketest.py
```diff
@@ -0,0 +1,51 @@
...
+        "layout-parser-paper-fast.jpg",
+    ]
+)
+
+def test_happy_path(example_filename):
+    """
+    For the files in sample-docs, verify that we get a 200
+    and some structured response
+    """
+    time.sleep(1)  # Avoid a 429
```
ryannikolaidis, line 40: for my own sake: what's the deal with the 429? I'm assuming this has to do with multiple back-to-back calls with the parameterization?

part 2: do we need the sleep at the beginning and end of the test?

scripts/docker-smoke-test.sh
```diff
@@ -0,0 +1,36 @@
...
+container_name=pipeline-general
+image_name=pipeline-family-general-dev:latest
+
+start_container() {
+    echo Starting container "$container_name"
+    docker run -p 8000:8000 -d --rm --name "$container_name" "$image_name" --port 8000 --host 0.0.0.0
+
+    # Wait until the api is ready
+    # We can be smarter here - wait until /healthcheck returns?
+    sleep 15
```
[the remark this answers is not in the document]
awalker4: Yeah, quick and easy check - might as well do it here

scripts/smoketest.py
awalker4: (Although this pr is getting hairier than I expected so maybe in a new branch)

ryannikolaidis: Ahhh, rate limiting, that makes sense now. 

Ohh nice! 🤞

Dockerfile
```diff
@@ -42,16 +49,16 @@ ENV PATH="/home/${NB_USER}/.local/bin:${PATH}"
...
-COPY exploration-notebooks exploration-notebooks
-COPY pipeline-notebooks pipeline-notebooks
-
 RUN python3.8 -m pip install pip==${PIP_VERSION} \
   && pip3.8 install  --no-cache  -r requirements-base.txt \
   && pip3.8 install --no-cache "detectron2@git+https://github.com/facebookresearch/detectron2.git@v0.6#egg=detectron2"
 
+COPY prepline_${PIPELINE_PACKAGE}/ prepline_${PIPELINE_PACKAGE}/
+COPY exploration-notebooks exploration-notebooks
+COPY pipeline-notebooks pipeline-notebooks
```
ryannikolaidis, line 58: good catch here.

Dockerfile
```diff
@@ -42,16 +49,16 @@ ENV PATH="/home/${NB_USER}/.local/bin:${PATH}"
...
+
 EXPOSE 5000
 
-ENTRYPOINT ["uvicorn", "unstructured_inference.api:app", \
+ENTRYPOINT ["uvicorn", "prepline_general.api.app:app", \
```
ryannikolaidis, line 62: This looks good now, how was this working before this change? Or was there a refactor that broke it at some point?

scripts/docker-smoke-test.sh
```diff
@@ -0,0 +1,52 @@
...
+    url=localhost:8000/healthcheck
+
+    for i in {1..5}; do
+        echo "$i" Waiting for response from "$url"...
+        if curl $url 2> /dev/null; then
+            echo
+            return
+        fi
+
+        sleep 5
```
ryannikolaidis, line 29: since this is only hitting it's own local test instance (we don't have to worry about DOSing ourselves, could we sleep shorter..say 1s?

```suggestion
        sleep 1
```

awalker4: Makes sense!

Dockerfile
awalker4: The makefile was overriding the entrypoint (note that unstructured-inference is doing this too so small thing to fix at some point).

scripts/docker-smoke-test.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+# docker-smoke-test.sh
+# Start the containerized api and run some end-to-end tests against it
+# There will be some overlap with just running a TestClient in the unit tests
+# Is there a good way to reuse code here?
+# Also note this can evolve into a generalized pipeline smoke test
+
+# shellcheck disable=SC2317  # Shellcheck complains that trap functions are unreachable...
```
ryannikolaidis, line 9: Oh, good call out. Given that it's only a couple of tests right now, feels okay-ish, but would be nice to have a pattern here if it makes sense and we're going to be doing this more often.

scripts/smoketest.py
ryannikolaidis: ohh yea, totally, and doesn't need to block this pr either if you'd rather cut after you get this in?

ryannikolaidis (approved):
This looks great! Nice work! One very small suggestion on that healthcheck.

Dockerfile
```diff
@@ -42,16 +49,16 @@ ENV PATH="/home/${NB_USER}/.local/bin:${PATH}"
...
   "--host", "0.0.0.0", \
   "--port", "5000"]
```
ryannikolaidis, line 64: why do we use port 5000 here by default?

PR #52, opened 2023-03-21
ci: test before publishing Docker images

ryannikolaidis:
Runs smoke test against the build Docker image that is published on pushes to main. Failing the smoke test prevents the image from being published.

Confirmed push by enabling on pr and testing run here: https://github.com/Unstructured-IO/unstructured-api/actions/runs/4481898966


awalker4 (approved):
Looks good!

PR #53, opened 2023-03-21
docs: update readme for public docker builds

ryannikolaidis:


README.md
```diff
@@ -51,6 +51,25 @@ It's also nice to show how to call the API function using pure Python.
 
 You can generate the FastAPI APIs from your pipeline notebooks by running `make generate-api`.
 
+## :dizzy: Instructions for using the Docker image
+
+The following instructions are intended to help you get up and running using Docker to interact with `unstructured-api`.
+See [here](https://docs.docker.com/get-docker/) if you don't already have docker installed on your machine.
+
+NOTE: the image is only supported for x86_64 hardware and known to have issues on Apple silicon.
```
awalker4, line 59: I'd note that this is specific to files that need to use inference, i.e. `.jpeg` and `.pdf`. When running the amd64 container on a mac, the behavior is an `unsupported hardware` error that causes the server to hang.

awalker4 (approved):

ryannikolaidis:
> Added a small update for supported filetypes. Note that we're missing some filetypes in the tests so I'll add to those before taking another pass at the list.

Awesome, thank you!

PR #65, opened 2023-03-31
Passing params to partition

amanda103:
Passing `file_filename` and `content_type` to `partition()` from pipeline api


pipeline-notebooks/pipeline-general.ipynb
```diff
@@ -540,19 +551,44 @@
...
     "    # NOTE(robinson) - This is a hacky solution due to\n",
     "    # limitations in the SpooledTemporaryFile wrapper.\n",
-    "    # Specifically, it does't have a `seekable` attribute,\n",
+    "    # Specifically, it doesn't have a `seekable` attribute,\n",
     "    # which is required for .pptx and .docx. See below\n",
     "    # the link below\n",
     "    # ref: https://stackoverflow.com/questions/47160211\n",
     "    # /why-doesnt-tempfile-spooledtemporaryfile-implement-readable-writable-seekable\n",
+    "    if file_content_type and file_content_type not in os.environ.get(\"UNSTRUCTURED_ALLOWED_MIMETYPES\"):\n",
+    "        raise UnsupportedFileError(message=f\"{file_content_type} not currently supported\")\n",
```
cragwolfe, line 581: this should already be taken care of by the unstructured-api-tools generated code, e.g. https://github.com/Unstructured-IO/unstructured-api-tools/blob/060ead4/test_unstructured_api_tools/pipeline-test-project/prepline_test_project/api/process_file_1.py#L60

pipeline-notebooks/pipeline-general.ipynb
```diff
@@ -540,19 +551,44 @@
...
+    "                    \"application/vnd.ms-excel,application/vnd.openxmlformats-officedocument.\" \\\n",
+    "                    \"presentationml.presentation,\" \\\n",
+    "                    \"application/vnd.ms-powerpoint,application/xml\"\n",
+    "\n",
+    "\n",
+    "if not os.environ.get(\"UNSTRUCTURED_ALLOWED_MIMETYPES\", None):\n",
+    "    os.environ[\"UNSTRUCTURED_ALLOWED_MIMETYPES\"] =  DEFAULT_MIMETYPES\n",
+    "\n",
+    "\n",
+    "def pipeline_api(file, filename='', file_content_type=None, file_filename=\"\", response_type=\"application/json\"):\n",
```
cragwolfe, line 572: this signature should not have `file_filename`.

This pipeline_api is primarily intended for API's, in which case `filename` means the name of the `file`, even though it later gets passed along as `file_filename` to partition. `pipeline_api` explicitly does not by convention support opening a local file referenced to by a filename (as `partition` does).

I.e., first three args/kwargs should match: https://github.com/Unstructured-IO/unstructured-api-tools/blob/060ead4/test_unstructured_api_tools/pipeline-test-project/prepline_test_project/api/process_file_1.py#L33

cragwolfe:
Couple of unittests should be added to validate the functionality in PR:

* in the successful case where partitioned json elements are returned, verify that the "filename" in one of those elements (or all) matches the POST'ed filename
* in the case where the filetype is not supported (e.g. an xml file), validate the response

test_general/api/test_app.py
```diff
@@ -57,4 +59,21 @@ def test_general_api(example_filename):
...
+@pytest.mark.parametrize(
+    "example_filename",
+    [
+        "fake-xml.xml",
+    ],
+)
+def test_general_api_raises_error(example_filename):
+    client = TestClient(app)
+    test_file = Path("sample-docs") / example_filename
+    with pytest.raises(ValueError):
```
cragwolfe, line 76: Hmm, if you added back:
```
class UnsupportedFileError(Exception):
    def __init__(self, message):
        self.message = message
        super().__init__(self.message)

DEFAULT_MIMETYPES = "application/pdf,application/msword,image/jpeg,image/png,text/markdown," \
                    "text/x-markdown,application/epub,application/epub+zip,text/html," \
                    "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet," \
                    "application/vnd.ms-excel,application/vnd.openxmlformats-officedocument." \
                    "presentationml.presentation," \
                    "application/vnd.ms-powerpoint,application/xml"


if not os.environ.get("UNSTRUCTURED_ALLOWED_MIMETYPES", None):
    os.environ["UNSTRUCTURED_ALLOWED_MIMETYPES"] =  DEFAULT_MIMETYPES
    
```
from earlier in this branch
to the pipeline-notebook and  rerun`make generate-api`
do we not get the ValueError?

(my prior comment had been about removing the block
```
    if file_content_type and file_content_type not in os.environ.get("UNSTRUCTURED_ALLOWED_MIMETYPES"):
        raise UnsupportedFileError(message=f"{file_content_type} not currently supported")
```
only inside of `pipeline_api` but keeping the rest)

pipeline-notebooks/pipeline-general.ipynb
```diff
@@ -540,20 +550,27 @@
...
-    "    # Specifically, it does't have a `seekable` attribute,\n",
+    "    # Specifically, it doesn't have a `seekable` attribute,\n",
     "    # which is required for .pptx and .docx. See below\n",
     "    # the link below\n",
     "    # ref: https://stackoverflow.com/questions/47160211\n",
     "    # /why-doesnt-tempfile-spooledtemporaryfile-implement-readable-writable-seekable\n",
     "    with tempfile.TemporaryDirectory() as tmpdir:\n",
-    "        _filename = os.path.join(tmpdir, filename.split('/')[-1])\n",
-    "        with open(_filename, \"wb\") as f:\n",
+    "        file_filename = os.path.join(tmpdir, filename.split('/')[-1])\n",
```
cragwolfe, line 563: this is causing the client's filename that was specified in their POST, e.g. `'sample-docs/alert.eml'` to get replaced with `alert.eml`, which then means the `metadata["filename"]` in elements later gets set to `alert.eml`, which is incorrect. we should honor the metadata `filename` the user included in their request.

pipeline-notebooks/pipeline-general.ipynb
```diff
@@ -540,20 +550,27 @@
...
+    "        with open(file_filename, \"wb\") as f:\n",
```
cragwolfe, line 564: not strictly related to this PR, but the hack to rewrite the posted `file` object to another file (which is non ideal for a couple of reasons, including performance) seems like it is only needed for `.pptx` and `.docx` per Robinson's comment. can we keep it to that case? (or address in other PR... just did not want to lose track of this)

test_general/api/test_app.py
amanda103: ohhh I see, I'll try that.

pipeline-notebooks/pipeline-general.ipynb
```diff
@@ -539,21 +548,45 @@
...
+    "                f.write(file.read())\n",
+    "            elements = partition(filename=_filename)\n",
+    "    else:\n",
+    "        try:\n",
+    "            elements = partition(file=file,\n",
+    "                                 file_filename=filename,\n",
+    "                                 content_type=file_content_type)\n",
+    "        except ValueError as e:\n",
+    "            if 'Invalid file' in e.args[0]:\n",
+    "                raise HTTPException(status_code=400, detail=f\"{file_content_type} not currently supported\")\n",
```
cragwolfe, line 588: otherwise, we should still `raise`

cragwolfe: (so the user would get a 500 Internal Server Error in this case)

cragwolfe (approved):
LGTM!

ryannikolaidis:
@amanda103 I'm pretty sure I have an answer for the failing tests. It looks like the issue comes up when we pass in a filename or file_filename and try to detect the filetype. The issue is that libmagic tries to open the file from the filename, which in the case of the docker container (and our remote api) doesn't exist on disk. 

So, I think what we should do in this particular case (filename but no file on disk) is defer to that next block (use the filename extension). I have a fix over here: https://github.com/Unstructured-IO/unstructured/pull/446. I'm pretty sure once this goes in we'll bump and bring the change in over here and it should _just work_ 🤞

alextkd:
With PDFs is still failing.. 
image

PR #66, opened 2023-04-01
docs: update README with notes about multi-platform

ryannikolaidis:
~Note: This is prepped for once we confirm this build via CI on main.~

Confirmed on both m1 and dev ec2 instance


cragwolfe (approved):
LGTM (once dependent PR merged and images validated)

PR #70, opened 2023-04-05
feat: add an m_strategy parameter to pipeline_api

ryannikolaidis:
* adds an `m_strategy` parameter with a string default of [“fast”] to `pipeline_api`
* adds a smoke test that tests the timing of a PDF with both strategies and verifies `fast` is _much_ faster than `hi_res`
* adds a unit test that validates a 400 error for any other value for `strategy`
* adds README mention of this functionality

## Testing

I ran the new smoke test on both the local mac and a remote dev instance. Both were clear the 4x threshold that we're testing against.

- Run the app (`make run-web-app` or preferred method that runs on localhost)
- Run the smoke test: `PYTHONPATH=. pytest scripts/smoketest.py::test_strategy_performance`

You can test any request with:
```
curl -X 'POST' \
  'http://localhost:8000/general/v0/general' \
  -F 'files=@sample-docs/layout-parser-paper.pdf' \
  -F 'strategy=hi_res' \ 
  | jq -C .
```
Currently you can see at a glance whether the hi_res or fast strategy is running by observing the output of the web-app. In the case of hi_res you will see progress bars and mention of `Loading the Tesseract OCR agent`

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ For example:
   | jq -C . | less -R
 ```
 
-We have two strategies for processing PDF files: `hi_res` and `fast`. `hi_res` takes a bit longer but provides better quality results. You can specify which you prefer with the `strategy` parameter.
+We have two strategies for processing PDF files: `hi_res` and `fast`. `hi_res` takes longer but provides better quality results. Conversely, `fast` is ideal for scenarios where time-to-result is the priority (`fast` is also the default)  You can specify which you prefer with the `strategy` parameter.
 
 ```
  curl -X 'POST' \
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -41,11 +41,22 @@ For example:
   'http://localhost:8000/general/v0/general' \
   -H 'accept: application/json' \
   -H 'Content-Type: multipart/form-data' \
-  -F 'files=@family-day.eml' \
+  -F 'files=@sample-docs/family-day.eml' \
+  | jq -C . | less -R
+```
+
+We have two strategies for processing PDF files: `hi_res` and `fast`. `hi_res` takes a bit longer but provides better quality results. You can specify which you prefer with the `strategy` parameter.
```
amanda103, line 48: maybe a note about default being 'fast'?

amanda103 (approved):
Nit on readme wording, otherwise lgtm!

README.md
ryannikolaidis: sounds good!

PR #71, opened 2023-04-05
chore: build with registry as cache

ryannikolaidis:
* updates docker-build.sh to use the image registry as a cache
* updates make commands to no longer explicitly set platform (but optionally can be assigned via env var)

## Testing

Previous CI docker-test run time: [32m](https://github.com/Unstructured-IO/unstructured-api/actions/runs/4616209135/jobs/8160967137) 
New CI docker-test run time: [7m20s](https://github.com/Unstructured-IO/unstructured-api/actions/runs/4623792027/jobs/8177994053?pr=71) 

- Run `make docker-build` and note that most of steps will say CACHED.
- Run `make docker-build` on amd vs arm and use `docker inspect image unstructured-api:dev | grep Architecture` and note that the respective image for your platform was built

Also, I triggered a publish flow to validate that this still works as expected. Here is that run: https://github.com/Unstructured-IO/unstructured-api/actions/runs/4631349496


Makefile
```diff
@@ -1,7 +1,7 @@
 PIPELINE_FAMILY := general
 PIPELINE_PACKAGE := general
 PACKAGE_NAME := prepline_${PIPELINE_PACKAGE}
-PIP_VERSION := 22.1.2
+PIP_VERSION := 22.2.1
```
cragwolfe, line 4: nit: i've been thinking to avoid pip bumps until we bump pip everywhere

Makefile
```diff
@@ -70,23 +70,25 @@ generate-api:
...
+
 .PHONY: docker-build
 docker-build:
-	PIP_VERSION=${PIP_VERSION} PIPELINE_FAMILY=${PIPELINE_FAMILY} PIPELINE_PACKAGE=${PIPELINE_PACKAGE} ./scripts/docker-build.sh
+	PIP_VERSION=${PIP_VERSION} PIPELINE_PACKAGE=${PIPELINE_PACKAGE} DOCKER_IMAGE_NAME=${DOCKER_IMAGE} ./scripts/docker-build.sh
 
 .PHONY: docker-start-api
 docker-start-api:
-	docker run -p 8000:8000 --mount type=bind,source=$(realpath .),target=/home/notebook-user/local -t --rm pipeline-family-${PIPELINE_FAMILY}-dev:latest --log-config logger_config.yaml --host 0.0.0.0 --port 8000
+	docker run -p 8000:8000 --mount type=bind,source=$(realpath .),target=/home/notebook-user/local -t --rm ${DOCKER_IMAGE} --log-config logger_config.yaml --host 0.0.0.0 --port 8000
```
cragwolfe, line 81: other nit: are these `make` targets used in the current CI process? if not, i'd lean towards leaving as is for consistency with other pipeline- repos.

more context: these make targets had been more oriented towards local dev in the past, and are basically templated outputs from the pipeline- standard (something created from https://github.com/Unstructured-IO/pipeline-template/ ).

that said, it's not a huge deal to stray from that -- bulk updates across pipeline-api's are still possible with the right scripting, just want to make sure.

cragwolfe:
> updates make commands to no longer explicitly set platform (but optionally can be assigned via env var)

seems like this is something to promote to pipeline-template and extend to other pipeline- repos.

A couple of nits, otherwise awesome improvement on CI! 🙌

Makefile
ryannikolaidis: ahh, I was bumping for consistency with [the unstructured repo](https://github.com/Unstructured-IO/unstructured/blob/ef9fb79ed482a5d2f8694fbbbcac3b81bbb1b85d/Makefile#L2).

but I am seeing that [other pipelines](https://github.com/Unstructured-IO/pipeline-sec-filings/blob/main/Makefile#L4) are still on 22.1.2. will revert

Makefile
ryannikolaidis: nope, not necessary for ci! was in here and doing some clean up,  also wasn't sure if we wanted to keep calling this image pipeline-general vs unstructured-api

I do think being able to specify/override the docker image build name by variable is pretty valuable, but to your point maybe a change we do across all of them.

ryannikolaidis:
> > updates make commands to no longer explicitly set platform (but optionally can be assigned via env var)
> 
> seems like this is something to promote to pipeline-template and extend to other pipeline- repos.

yep, will do!

Makefile
cragwolfe: yeah, that's a good point, and probably worth generalizing this across the other pipeline- repos. just need to remember to do that :)

cragwolfe (approved):

Makefile
ryannikolaidis: cool! will add that to a the ticket I'm writing for the platform variable and we can knock both out at once.

PR #81, opened 2023-04-11
feat: Add coordinates param to api

awalker4:
Coordinates are no longer returned by default in elements. Pass `coordinates=true` to recover this functionality.


ryannikolaidis (approved):
since this changes how it currently works, worth a bump in the README or any other documentation? 

otherwise, looks great to me! tested by running `make run-web-app` and then curling with and without `-F 'coordinates=true'`.

awalker4:
Ah, good call! Let me know what you think of the new section in the readme.

ryannikolaidis:
> Ah, good call! Let me know what you think of the new section in the readme.

Perfect!

PR #84, opened 2023-04-13
Update api-tools version and fix openapi_url

awalker4:
This is to grab the change from https://github.com/Unstructured-IO/unstructured-api-tools/pull/164 and ensure we serve the api spec at /general/openapi.json.


cragwolfe (approved):
🎉

Issue #83, opened 2023-04-11
Posting a bad .pdf results in a 500


cragwolfe:
A file that either has the .pdf extension or has the `Content-Type: application/pdf` multipart/form header that is not truly a .pdf file (such as plain text), results in a 500 with a stack trace, shown below.

Instead, a 400 should be returned with a message indicating the issue. This should be the case for `strategy=fast` or `strategy=hi_res`.

In the future, the app should be able to process a payload based on inspection rather than filename (or Content-Type: multipart/form header) given the presence of another form parameter (e.g., `infer-content-type-from-file`).

```
  File "/Users/cragwolfe/r/unstructured-api/prepline_general/api/general.py", line 291, in pipeline_1
    response = pipeline_api(
  File "/Users/cragwolfe/r/unstructured-api/prepline_general/api/general.py", line 98, in pipeline_api
    elements = partition(
  File "/Users/cragwolfe/.pyenv/versions/unstapi/lib/python3.8/site-packages/unstructured/partition/auto.py", line 82, in partition
    return partition_pdf(
  File "/Users/cragwolfe/.pyenv/versions/unstapi/lib/python3.8/site-packages/unstructured/partition/pdf.py", line 50, in partition_pdf
    return partition_pdf_or_image(
  File "/Users/cragwolfe/.pyenv/versions/unstapi/lib/python3.8/site-packages/unstructured/partition/pdf.py", line 111, in partition_pdf_or_image
    return _partition_pdf_with_pdfminer(
  File "/Users/cragwolfe/.pyenv/versions/unstapi/lib/python3.8/site-packages/unstructured/utils.py", line 40, in wrapper
    return func(*args, **kwargs)
  File "/Users/cragwolfe/.pyenv/versions/unstapi/lib/python3.8/site-packages/unstructured/partition/pdf.py", line 213, in _partition_pdf_with_pdfminer
    elements = _process_pdfminer_pages(
  File "/Users/cragwolfe/.pyenv/versions/unstapi/lib/python3.8/site-packages/unstructured/partition/pdf.py", line 240, in _process_pdfminer_pages
    for i, page in enumerate(PDFPage.get_pages(fp)):
  File "/Users/cragwolfe/.pyenv/versions/unstapi/lib/python3.8/site-packages/pdfminer/pdfpage.py", line 151, in get_pages
    doc = PDFDocument(parser, password=password, caching=caching)
[...]

PR #88, opened 2023-04-19
fix: malformed .pdf no longer results in a 500

natygyoon:
https://github.com/Unstructured-IO/unstructured-api/issues/83#issue-1663388581


cragwolfe (approved):
LGTM!

Issue #85, opened 2023-04-14
Ability to return text/csv instead of json


cragwolfe:
The pipeline_api should be able to return a text/csv response instead of json if the [response_type passed through pipeline_api](https://github.com/Unstructured-IO/unstructured-api/blame/186db764218da8be2046bd5bce732196181d7ef6/pipeline-notebooks/pipeline-general.ipynb#L565) is "text/csv".

In this case, the the pipeline_api should call [convert_to_csv](https://github.com/Unstructured-IO/unstructured/blob/3972c80c518fb3626a9437709cbc84a3eb85c47d/unstructured/staging/base.py#L137) before returning the result.

**Definition of Done**

* Unittests adding confirming the new functionality
* Plus a [smoketest](https://github.com/Unstructured-IO/unstructured-api/blob/main/scripts/smoketest.py)
* README is updated indicating this capability

PR #100, opened 2023-05-08
Feat: Return text/csv instead json

kravetsmic:
Updated:
* pipeline_api: added returning `text/csv` instead `json`, if `response_type="text/csv"`
* smoketests: added additional tests for testing returning `text/csv` response
* API: updated API for new changes

Closes issue #85 

---

Bump version to 0.0.20


scripts/smoketest.py
```diff
@@ -1,56 +1,114 @@
...
 
-def send_document(filename, content_type, strategy="fast"):
+def send_document(filename, content_type, strategy="fast", output_format="application/json"):
     files = {"files": (str(filename), open(filename, "rb"))}
-    return requests.post(API_URL, files=files, data={"strategy": strategy})
+    return requests.post(API_URL, files=files, data={"strategy": strategy, "output_format": output_format})
 
 
+# NOTE(kravetsmic): This function was added for removing path from text/csv response because the backend has a different
+# temporary path than the temporary path in tests
```
cragwolfe, line 22: the backend server should never be sending back a temporary path (unless the client actually posted that as the filename in its form submission as the filename)

cragwolfe: so, there is another bug that needs to be resolved here, not related to this PR

cragwolfe: i believe the bug that prompted the workaround here has now been resolved in main, so hopefully this `remove_path` function may be removed.

cragwolfe: @kravetsmic  , is this note still valid? any temporary paths should have been removed with https://github.com/Unstructured-IO/unstructured-api/commit/d876a86195ed94e7fcd1d5b53662ab5aa15d27e1#diff-d80162975e0983c4067dc93f0d9a06f8867a7a982cf67a38975109e448d20c5e

scripts/smoketest.py
```diff
@@ -1,56 +1,101 @@
...
+        filename=test_file,
+        strategy=strategy,
+        content_type=content_type,
+        output_format=output_format,
+    )
 
     assert(response.status_code == 200)
     assert len(response.json()) > 0
-    assert len("".join(elem["text"] for elem in response.json())) > 20
+    # NOTE(kravetsmic): looks like a bug on macOS (m1), incorrectly scanned text from images and pdf files
```
cragwolfe, line 78: please say more. is it all images and pdf's or are there specific files causing a problem? this may be worth creating an issue for in the unstructured repo.

cragwolfe:
Closing since resolved by https://github.com/Unstructured-IO/unstructured-api/pull/125 .

PR #103, opened 2023-05-10
chore: remove docx and pptx writing to file

ryannikolaidis:
Since we now push docx and pptx files into BytesIO in unstructured we no longer need to write them to a temporary file in the pipeline.

Also we no longer need to attempt to downgrade urllib since we have openssl 1.1.1 install in the image build. Removes downgrade step.

* Undo special handling of docx and pptx getting written to temp file
* bonus: Undo revertting to urllib<2

### DIFF
--- a/prepline_general/api/general.py
+++ b/prepline_general/api/general.py
@@ -73,34 +73,18 @@ def pipeline_api(
     show_coordinates_str = (m_coordinates[0] if len(m_coordinates) else "false").lower()
     show_coordinates = show_coordinates_str == "true"
 
-    if filename.endswith((".docx", ".pptx")):
-        # NOTE(robinson) - This is a hacky solution due to
-        # limitations in the SpooledTemporaryFile wrapper.
-        # Specifically, it doesn't have a `seekable` attribute,
-        # which is required for .pptx and .docx. See below
-        # the link below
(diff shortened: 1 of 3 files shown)


tabossert (approved):

prepline_general/api/general.py
```diff
@@ -73,34 +73,18 @@ def pipeline_api(
...
-        except pdfminer.pdfparser.PDFSyntaxError:
+    try:
+        elements = partition(
+            file=file, file_filename=filename, content_type=file_content_type, strategy=strategy
+        )
+    except ValueError as e:
+        if "Invalid file" in e.args[0]:
             raise HTTPException(
-                status_code=400, detail=f"{filename} does not appear to be a valid PDF"
+                status_code=400, detail=f"{file_content_type} not currently supported"
```
cragwolfe, line 83: probably worth a ticket to not make such a broad claim with valueerror, but not related to this ticket

cragwolfe (approved):

prepline_general/api/general.py
ryannikolaidis: 100% will create that ticket

cragwolfe: thanks!

PR #105, opened 2023-05-19
fix: EXPOSE default port

dkarlovi:
The network port must be marked as exposed for Docker to know the container has network connectivity by default.


awalker4:
Thanks for the pr! If I recall, we recently removed an EXPOSE line from the dockerfile. @ryannikolaidis what's the history on that?

ryannikolaidis:
> Thanks for the pr! If I recall, we recently removed an EXPOSE line from the dockerfile. @ryannikolaidis what's the history on that?

@dkarlovi are you running into this issue despite running docker with `-p 8000:8000 --port 8000`? Or are you trying to run without specifying these? Is there a specific use-case? We ran into cases where we and users may want to specify ports other than 8000 and where it felt misleading to expose other ports by default.

dkarlovi:
> @dkarlovi are you running into this issue despite running docker with `-p 8000:8000 --port 8000`? Or are you trying to run without specifying these? Is there a specific use-case?

No, I actually don't want to bind to any ports. The issue here is mostly that some tooling will inspect containers running and run additional work if they have networking support declared. Without `EXPOSE` the container is by default seen as "doesn't have networking".

I've fixed it for myself by adding `--expose` (this is not the same as `--port`).

> We ran into cases where we and users may want to specify ports other than 8000 and where it felt misleading to expose other ports by default.

I guessed as much, if you provide a way to redefine the port, it then makes sense to not define this. But since the image runs with a default port, it should by default also expose it IMO. The value of having the port be dynamic *inside* the container is IMO limited, even if you want to change it you can just do `--port 1234:8000`.

ryannikolaidis:
> > @dkarlovi are you running into this issue despite running docker with `-p 8000:8000 --port 8000`? Or are you trying to run without specifying these? Is there a specific use-case?
> 
> No, I actually don't want to bind to any ports. The issue here is mostly that some tooling will inspect containers running and run additional work if they have networking support declared. Without `EXPOSE` the container is by default seen as "doesn't have networking".
> 
> I've fixed it for myself by adding `--expose` (this is not the same as `--port`).
> 

Ahh okay. Yep, that was where I was headed. I wasn't meaning to conflate expose vs port, but noticed folks often do so wanted to clarify precisely your need here.

> > We ran into cases where we and users may want to specify ports other than 8000 and where it felt misleading to expose other ports by default.
> 
> I guessed as much, if you provide a way to redefine the port, it then makes sense to not define this. But since the image runs with a default port, it should by default also expose it IMO. The value of having the port be dynamic _inside_ the container is IMO limited, even if you want to change it you can just do `--port 1234:8000`.

Yea, definitely a very fair point on defaults. I guess the one awkward bit in my mind is that afaict if you expose the port in the image there's no way to un-expose if you use decide to use a different port elsewhere. Whereas, to your point, a user can always  just explicitly call with `--expose`.

How would you feel if instead we add `--expose 8000` to the default entrypoint in the Dockerfile? It seems like that would address your point that defaults should _just work_ out of the box (totally fair), but also is easy to explicitly override.

dkarlovi:
> Yea, definitely a very fair point on defaults. I guess the one awkward bit in my mind is that afaict if you expose the port in the image there's no way to un-expose if you use decide to use a different port elsewhere. Whereas, to your point, a user can always just explicitly call with `--expose`.

That is true, but `EXPOSE` doesn't really do anything by default, it's almost a metadata hint which says
> This image has networking on port X.

When the container runs, it doesn't need to be true. Some images mark many ports exposed (one example I've seen recently is Jaeger, MySQL does too).

> How would you feel if instead we add `--expose 8000` to the default entrypoint in the Dockerfile? It seems like that would address your point that defaults should _just work_ out of the box (totally fair), but also is easy to explicitly override.

IIUC, you're suggesting to run `--expose` *in* the image's entrypoint? You can't do that AFAIK, the entrypoint runs within the container, once the container is running, the phase where EXPOSE is relevant has past.

ryannikolaidis:
> That is true, but `EXPOSE` doesn't really do anything by default, it's almost a metadata hint which says

Yea, that's kinda where my head was. I guess fine if we want to do this, but maybe drop an extra comment on that line for context?

dkarlovi:
Sure, feel free to suggest some wording. :+1:

Dockerfile
```diff
@@ -120,3 +120,4 @@ COPY --chown=${NB_USER}:${NB_USER} pipeline-notebooks pipeline-notebooks
 ENTRYPOINT ["uvicorn", "prepline_general.api.app:app", \
   "--log-config", "logger_config.yaml", \
   "--host", "0.0.0.0"]
+EXPOSE 8000
```
ryannikolaidis, line 123:
```suggestion
# Expose a default port of 8000. Note: The EXPOSE instruction does not actually publish the port, 
# but some tooling will inspect containers and perform work contingent on networking support declared. 
EXPOSE 8000
```

ryannikolaidis (approved):
lgtm! thanks for posting this but also taking the time and talking it through, really appreciate it!

dkarlovi:
Thanks for your help @ryannikolaidis

PR #108, opened 2023-05-25
chore: fixing cors var handling

amanda103:
Bumping the version and running make generate-api to alter optional os env var ALLOWED_ORIGINS. The env var comes in as a comma separated string and we need a list to pass to the FastAPI CORS implementation (ex: `"http://example.com,http://anotherdomain.com"`)

https://github.com/Unstructured-IO/unstructured-api-tools/pull/174


awalker4:
LGTM!

awalker4 (approved):

PR #109, opened 2023-05-26
Chore: sync with new pandoc version on new base image 

yuming-long:
### Summary

Add tests for more filetypes (epub, odt, rft) in smocktest to sync with newer pandoc version in base image. 
Will publish this commit to both dev and prod.


awalker4 (approved):
LGTM! Quick note - can you add these to the supported file types at the top of the readme?

PR #115, opened 2023-06-05
feat: Add pdf_processing_mode to enable running pdfs in parallel

awalker4:
This is a quick win for parallel processing of pdfs. Adds a new `pdf_processing_mode` param to the api. Default is `serial`, but when the mode is `parallel`, we'll split pdfs into 1 page files and send each to the hosted api.

With our `layout-paser-paper.pdf`, I'm seeing a ~3x speedup when running `hi_res`. This clearly isn't the best we can do, but when the full file is taking ~90 seconds, we have to start somewhere.

## Limitations / things to note:
- The upstream url is hardcoded to api.unstructured.io. This should really be whatever url is used to access the initial server, which we can get from `Request.url`. This is a `api-tools` change to pass the Request object down.
- The hosted api is rate limited to 100 requests per 5 minutes. Considering that the 1 pdf above will fan out to 16 additional requests, we need to bump this. Alternatively, we can whitelist our ECS hosts to allow traffic from ourselves to exceed this. Figure this out before this is deployed.
- This PR exposes some limitations in `partition_via_api`:
  -  There's a bug where the element metadata has filetype json
  -  Errors from the upstream api look like this: `Received unexpected status code 400 from the API`. We should make it easier to propagate the status code, plus any details the api returned.

## Testing
Please try something other than `layout-paper-parser.pdf` to verify the response times. First, try your pdf against our api for `hi_res`:

```
curl 'https://api.unstructured.io/general/v0/general' \
--header 'Accept: application/json' \
--form 'files=@"~/test files/layout-parser-paper.pdf"' \
--form 'strategy="hi_res"'
```

Then, pull this branch and run `make run-web-app`. Try the same pdf locally, which will hit the hosted api in parallel (page size should be under 100 per the rate limiting).

```
curl --location 'http://localhost:8000/general/v0/general' \
--header 'Accept: application/json' \
--form 'files=@"~/test files/layout-parser-paper.pdf"' \
--form 'strategy="hi_res"' \
--form 'pdf_processing_mode="parallel"'
```

Do this with cli timing tool, or via postman, to confirm that you're seeing a speedup.

### DIFF
--- a/prepline_general/api/general.py
+++ b/prepline_general/api/general.py
@@ -17,6 +17,10 @@ from starlette.types import Send
 from base64 import b64encode
 from typing import Optional, Mapping, Iterator, Tuple
 import secrets
+from concurrent.futures import ThreadPoolExecutor
+from functools import partial
+from PyPDF2 import PdfReader, PdfWriter
+from unstructured.partition.api import partition_via_api
 from unstructured.partition.auto import partition
 from unstructured.staging.base import convert_to_isd
 import tempfile
@@ -34,9 +38,6 @@ def is_expected_response_type(media_type, response_type):
         return True
     else:
         return False
-
-
-# pipeline-api
 
 
 DEFAULT_MIMETYPES = (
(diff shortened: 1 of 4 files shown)


ryannikolaidis:
[looks like](https://github.com/Unstructured-IO/unstructured-api/actions/runs/5181919918/jobs/9338130347?pr=115) you need to add PyPDF2 as a test dependency and pip compile

test_general/api/test_app.py
```diff
@@ -144,3 +146,82 @@ def test_general_api_returns_500_bad_pdf():
     assert response.json() == {"detail": f"{tmp.name} does not appear to be a valid PDF"}
     assert response.status_code == 400
     tmp.close()
+
+
+def test_parallel_mode_correct_result():
+    """
+    Validate that splitting up the pages and sending to the api
+    returns the same result as usual
+    Note that this test makes calls to prod for now
```
ryannikolaidis, line 155: Not a huge fan of tests that hit prod. Are we doing this elsewhere? Any alternatives?

prepline_general/api/general.py
```diff
@@ -58,28 +59,115 @@ def is_expected_response_type(media_type, response_type):
...
     filename="",
     m_strategy=[],
     m_coordinates=[],
+    m_pdf_processing_mode=[],
     file_content_type=None,
     response_type="application/json",
 ):
     strategy = (m_strategy[0] if len(m_strategy) else "fast").lower()
-    if strategy not in ["fast", "hi_res"]:
+    strategies = ["fast", "hi_res", "auto"]
```
ryannikolaidis, line 143: good catch on "auto"!

ryannikolaidis: wonder (longterm, not this PR) if these should be defined in a way that doesn't result in loose lists of strings like this. i.e. enum or something alongside original definition that forces us to conform.

prepline_general/api/general.py
```diff
@@ -58,28 +59,115 @@ def is_expected_response_type(media_type, response_type):
...
+
+def partition_file_via_api(file, **kwargs):
+    """
+    Given a file-like, use partition_via_api to retrieve its elements.
+    Hardcode sending it to api.unstructured.io for now.
+    """
+    # Note (austin) - the correct thing is to call back to ourselves
+    # using the request.url. We'll need to pass the Request into pipeline_api().
+    # Same goes for any request context to pass on, i.e. the token
+    request_url = "https://api.unstructured.io/general/v0/general"
```
ryannikolaidis, line 96: any reason not to do that now? some unexpected complexity?

awalker4: Mostly just that it's an api-tools change. We'll be back in here to figure out how to improve the speedup, so I figured I'd get something out to start.

test_general/api/test_app.py
awalker4: Hmm, to your other point, I guess we can override the url in this case to point it back to itself. I considered mocking it at first, but we want to validate the results, so it feels artificial to be testing against whatever value we put in there.

awalker4:
Oh wait, I totally should have a env variable to set the mode. This shouldn't be set at the request level... I'm thinking it's disabled by default - no one needs this added complexity running locally. Does that make sense?

prepline_general/api/general.py
cragwolfe: the request_url should:

1/ be an env var
2/ not be the external DNS name in the first place (i.e., you probably don't want to route through the internet)

awalker4: Ah, good point! I had also been wondering if it's worth having a private load balancer for this traffic, but that's perhaps too much infra work right now.

test_general/api/test_app.py
awalker4: Okay, I've mocked it to switch out the api call for `partition()` which allows us to test merging the results without dependencies.

ryannikolaidis: Awesome, thanks. Yea, this makes sense to me!

ryannikolaidis: Awesome, thanks. Yea, this makes sense to me!

test_general/api/test_app.py
```diff
@@ -144,3 +147,89 @@ def test_general_api_returns_500_bad_pdf():
...
+        MAIN_API_ROUTE,
+        files=[("files", (str(test_file), open(test_file, "rb"), "application/pdf"))],
+        data={"pdf_processing_mode": "parallel"},
+    )
+
+    assert response.status_code == 500
+
+    # TODO (austin) - Right now any non 200 is going to turn into a 500
+    # because of how partition_via_api_works
+    # At the very least we can return a message with a bit more info
```
ryannikolaidis, line 220: is there a ticket tracking on what needs to change for this? Assuming ideally we just pass along the same non-200 status code, right? Can we pull from: https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/api.py#L180 or maybe a bump to that code to raise a custom exception where we assign the original status code explicitly?

Not a blocker for this, just want to make sure we're tracking.

ryannikolaidis (approved):
Thanks for the changes, this looks good to me! Dropped a couple of questions, but nothing that needs to block, esp since this is disabled by default. Worth hammering this a bit in dev.

test_general/api/test_app.py
awalker4: Ticket now made [here](https://unstructured-ai.atlassian.net/browse/CORE-1220). Yeah, I think it makes sense to pull the status code out of that string. It's worth a bump to that code to be able to return any details from the upstream call, though.

prepline_general/api/general.py
```diff
@@ -106,6 +108,10 @@ def partition_file_via_api(file, **kwargs):
         # Need to pull that back out and set it here
         raise HTTPException(status_code=500, detail=f"{e}")
 
+    # We need to account for the original page numbers
+    for element in elements:
+        element.metadata.page_number += page_offset
+
```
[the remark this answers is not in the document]
awalker4: Thanks to the `unstructured` fix to bring back page numbers, this actually got caught by `test_parallel_mode_correct_results`! Which also means the mocking I added is doing the right thing 🎉

test_general/api/test_app.py
ryannikolaidis: yep, completely agree. thanks!

prepline_general/api/general.py
```diff
@@ -58,6 +62,84 @@ def is_expected_response_type(media_type, response_type):
...
+    if not request_url:
+        raise HTTPException(status_code=500, detail="Parallel mode enabled but no url set!")
+
+    file, page_offset = file_tuple
+
+    try:
+        # Note (austin) - consider using partition_multiple_via_api to cut down on traffic
+        # This would serialize the work that we just tried to split up,
+        # so we first need to parallelize multi file input
+        elements = partition_via_api(file=file, api_url=request_url, **kwargs)
```
cragwolfe, line 108: might want a retry decorator here. ie, retry once on a spurious error

awalker4: That makes sense! I'll make a note for the next pass.

PR #122, opened 2023-06-09
Various updates to parallel mode

awalker4:
* Replace partition_via_api with a requests.post, and pass on any relevant headers using the FastAPI request.
* add smoke test to compare parallel and single mode.
* xfail jpeg in tests until we fix the strategy=fast issue - https://github.com/Unstructured-IO/unstructured/pull/735 (**Nevermind**)

To verify, the test instructions from https://github.com/Unstructured-IO/unstructured-api/pull/115 still apply.


pipeline-notebooks/pipeline-general.ipynb
```diff
@@ -608,32 +609,33 @@
     "    return split_pdfs\n",
     "\n",
     "\n",
-    "def partition_file_via_api(file_tuple, **kwargs):\n",
+    "def partition_file_via_api(file_tuple, request, filename, content_type, **request_kwargs):\n",
```
yuming-long, line 612: i am not sure if `request_kwargs` is the same thing as `request_data_kwargs` i put in https://github.com/Unstructured-IO/unstructured/pull/716/files#diff-1904aaa4e2be0c8efe726dbc2614b4080a6074d816d249cfbaddc1e1ab8738a0R22, if it is i think its better if we can sync them? I can change mine back to `request_kwargs`

awalker4: Hmm, yeah, would make sense to have them consistent. I think request_args is a bit more succint.

scripts/docker-smoke-test.sh
```diff
@@ -38,18 +48,42 @@ await_server_ready() {
...
+#######################
+echo Running parallel mode test
+
+curl http://localhost:$API_PORT/general/v0/general --form 'files=@sample-docs/layout-parser-paper.pdf' --form 'coordinates=true' --form 'strategy=fast' | jq -S > single-mode.json
+
+stop_container
+start_container "true"
+await_server_ready
+
+curl http://localhost:$API_PORT/general/v0/general --form 'files=@sample-docs/layout-parser-paper.pdf' --form 'coordinates=true' --form 'strategy=fast' | jq -S > parallel-mode.json
```
cragwolfe, line 79: also test `strategy=hi_res` ?

pipeline-notebooks/pipeline-general.ipynb
```diff
@@ -685,6 +705,11 @@
     "            detail=f\"Invalid strategy: {strategy}. Must be one of {strategies}\"\n",
     "        )\n",
     "        \n",
+    "    # Note (austin) - there is no fast mode for images\n",
+    "    # In this case we need to use hi_res\n",
+    "    if file_content_type == \"image/jpeg\":\n",
+    "        strategy = \"hi_res\"\n",
```
cragwolfe, line 711: this should probably be handled at the unstructured `partition` level

prepline_general/api/general.py
```diff
@@ -129,11 +133,21 @@ def partition_pdf_splits(file, **kwargs):
...
     pages = get_pdf_splits(pdf, split_size=pages_per_pdf)
-    partition_func = partial(partition_file_via_api, **kwargs)
+
+    partition_func = partial(
+        partition_file_via_api,
+        request=request,
+        filename=file_filename,
+        content_type=content_type,
+        coordinates=coordinates,
+        **request_kwargs,
```
cragwolfe, line 149: what would be in `request_kwargs` ?

prepline_general/api/general.py
```diff
@@ -90,28 +91,31 @@ def get_pdf_splits(pdf, split_size=1):
     return split_pdfs
 
 
-def partition_file_via_api(file_tuple, **kwargs):
+def partition_file_via_api(file_tuple, request, filename, content_type, **request_kwargs):
```
cragwolfe, line 94: maybe `request_kwargs` could be renamed to `form_parameter_kwargs`? this only pertains to form parameters, right?

The docstring below should be updated, might as will include comments for the kwargs there.

scripts/docker-smoke-test.sh
cragwolfe: another test for `coordinates=false`, or not provided?

pipeline-notebooks/pipeline-general.ipynb
yuming-long: i can bump unstructured again, but when i was testing the ocr parameters i noticed `hi_res` does not work with `ocr_languages` parameter. so maybe we should covert strategy to `auto` if a `fast` strategy is passed in for images in unstructured?

cragwolfe: i think `fast` -> `ocr_only` for images makes sense, but this should happen in the unstructured lib.

> i noticed hi_res does not work with ocr_languages parameter. 

er, is this a bug for unstructured itself? seems like it should.

yuming-long: l i think its expected: https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/strategies.py#L127-L137

scripts/docker-smoke-test.sh
awalker4: Ah, good call. I'll have this test loop through a number of curl commands.

pipeline-notebooks/pipeline-general.ipynb
awalker4: For now I'll disable the failing test on the api side

prepline_general/api/general.py
awalker4: Good catch - updated the docstrings. I renamed this to `partition_kwargs` to clarify that these are parameters that will be passed on, or sent to local partition. This allows me to exclude things like `coordinates`, which will be forwarded, but not passed to local partition.

prepline_general/api/general.py
```diff
@@ -120,33 +130,55 @@ def partition_file_via_api(file_tuple, **kwargs):
...
     return results
 
 
 def pipeline_api(
     file,
+    request=None,
     filename="",
     m_strategy=[],
     m_coordinates=[],
+    m_use_parallel_mode=[],
```
cragwolfe, line 181: Ah, we should not be exposing this to end users. In other words, an end (external) user should not be able to determine the behavior by setting this.

cragwolfe (changes requested):

prepline_general/api/general.py
awalker4: Makes sense! I was looking for a good way to run tests against parallel mode. For the new test script, I suppose we can start two containers side by side, but without a runtime parameter we wouldn't be able to compare parallel/non-parallel responses against our hosted api without a manual restart. I'll make a note of that in the test script.

cragwolfe: yeah, i was assuming we'd just restart the web server to test consistency between the two modes.

awalker4:
@cragwolfe when you have a moment can you take another look?

cragwolfe (approved):
Not tested, but do not see any issues.
Did a particular version of unstructured-api-tools allow you to pass in `request` to `pipeline_api` correctly?

awalker4:
Yep! Added it as an optional param here: https://github.com/Unstructured-IO/unstructured-api-tools/pull/175

Issue #123, opened 2023-06-12
Question: Is there a way to use the ocr_languages parameter with the API


demux79:
Images and PDF support the ocr_languages parameter in unstructured. Is this supported by the API as well?

yuming-long:
Hi @demux79, Thank you for bring this issue! I've merged a PR for this but haven't deploy the image to our API yet. For now you could use `http://localhost:8000/general/v0/general` with ocr parameters. I will let you know once i deployed it.

yuming-long:
okay its deployed to `https://api.unstructured.io/general/v0/general` 🙏

PR #124, opened 2023-06-12
Chore: Adds support for `ocr_only` and `ocr_languages`

yuming-long:
Draft for now, need a better readme

### Summary
Closes https://github.com/Unstructured-IO/unstructured-api/issues/123 and adds supports for `ocr_only` startegy and `ocr_languages` parameters

### Test
* `make docker-start-api` or `make run-web-app`
* curl with
```
curl  -X 'POST'  'http://localhost:8000/general/v0/general'  \
 -F 'files=@sample-docs/english-and-korean.png' \
-F 'strategy=ocr_only' \
 -F 'ocr_languages=eng'  \
 -F 'ocr_languages=kor'  \
 | jq -C . | less -R 
 ```

### CODE REVIEW

File: prepline_general/api/general.py [hunk 1]
```diff
@@ -147,31 +147,47 @@ def pipeline_api(
...
+    strategies = ["fast", "hi_res", "auto", "ocr_only"]
     if strategy not in strategies:
         raise HTTPException(
             status_code=400, detail=f"Invalid strategy: {strategy}. Must be one of {strategies}"
         )
 
+    # Note (austin) - there is no fast mode for images
+    # In this case we need to use hi_res
+    if file_content_type in ["image/jpeg", "image/png"]:
+        strategy = "hi_res"
```
cragwolfe, line 164: since the awkwardness of passing an irrelevant strategy for an image file type exists in the library itself in [partition()](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py), i think this kind of logic should live there, not in the API.

another reason is that unstructured-ingest would have the same issue using a single `strategy` across many doc types.

cragwolfe (approved):

PR #125, opened 2023-06-13
Allow for CSV response

rbiseck3:
Added support for a csv/text response rather than json if response type set to `text/csv`


rbiseck3:
Tackling issue: [#85](https://github.com/Unstructured-IO/unstructured-api/issues/85)

awalker4 (commented):
Looks good! I think to avoid the big parameter copy we should add the csv tests as an extra POST in test_general_api, etc.

cragwolfe (approved):
Looks good, but please update the README as well.

PR #126, opened 2023-06-16
chore: bump python to 3.8.17

cragwolfe:
The images pushed to quay.io will now have python 3.8.17 rather than python 3.8.15.


awalker4 (approved):
Hooray!

PR #127, opened 2023-06-19
Chore: Update doc and tests for filetypes

yuming-long:
### Summary

* Check the parameterized tests represent all types supported by [auto partition](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/auto.py#L14).
* Add support for .rst, .tsv, .xml files


scripts/install-pandoc.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
```
cragwolfe, line 2: i'd check if `uname` == "darwin" (the bash appropriate way here) and exit if that is the case with a friendly message saying only intended for linux.

awalker4 (approved):

PR #139, opened 2023-06-30
Feat: add `pdf_infer_table_structure` parameter to api

yuming-long:
### Summary

Add param `pdf_infer_table_structure` to  include support for tables. `pdf_infer_table_structure` is set to `False` by default

### Tests
* `make docker-build`
* Run `make docker-start-api`
* Test with 
```
curl -X 'POST'  'http://localhost:8000/general/v0/general' \
 -F 'files=@sample-docs/layout-parser-paper.pdf'  \
 -F 'strategy=hi_res' \
 -F 'pdf_infer_table_structure=true' \
| jq -C . | less -R
```
and table structure is included in the output:
Screenshot 2023-06-05 at 3 45 01 PM
* set `pdf_infer_table_structure=false` and table structure is not included:
Screenshot 2023-06-05 at 3 39 44 PM


pipeline-notebooks/pipeline-general.ipynb
```diff
@@ -726,6 +726,7 @@
     "    m_ocr_languages=[],\n",
     "    m_encoding=[],\n",
     "    m_xml_keep_tags=[],\n",
+    "    m_pdf_infer_table_structure = [],\n",
```
rbiseck3, line 729: Why is this an array?

yuming-long: this is how we pass in self defined parameters in api tools repo. It can support multiple input like this example:
```
curl -X 'POST' \
  'https://api.unstructured.io/general/v0/general' \
  -F 'files=@sample-docs/english-and-korean.png' \
  -F 'strategy=ocr_only' \
  -F 'ocr_languages=eng'  \
  -F 'ocr_languages=kor'  \
  | jq -C . | less -R
  ```
  so `m_ocr_languages=["eng", "kor"]` is  passing into the function this way (list of str)

Comment (rbiseck3) [reply pipeline-notebooks/pipeline-general.ipynb:729]: What potential values would be provided for the `pdf_infer_table_structure` parameter? I thought it was just a boolean.

Comment (yuming-long) [reply pipeline-notebooks/pipeline-general.ipynb:729]: yep it is one boolean value, the value is obtained here: https://github.com/Unstructured-IO/unstructured-api/pull/139/files#diff-d80162975e0983c4067dc93f0d9a06f8867a7a982cf67a38975109e448d20c5eR217

Comment (cragwolfe) [reply pipeline-notebooks/pipeline-general.ipynb:729]: yeah, more context is unstructured-api-tools only allow for string array params right now, so that's why everything has the default array value

Comment (rbiseck3) [reply pipeline-notebooks/pipeline-general.ipynb:729]: Ah that's what I was missing and explains this. I didn't realize boolean values weren't getting converted at the moment. Thanks for that info!

**Comment by shreyanid** [comment]
I don't see any functionality tests for this new param in `test_general/api/test_app.py` - could you add a test that the table structure data is only included when the parameter is set to true? (similar to the one in your description)

File: pipeline-notebooks/pipeline-general.ipynb [hunk 2]
```diff
@@ -756,6 +757,14 @@
     "    xml_keep_tags_str = (m_xml_keep_tags[0] if len(m_xml_keep_tags) else \"false\").lower()\n",
     "    xml_keep_tags = xml_keep_tags_str == \"true\"\n",
     "    \n",
+    "    pdf_infer_table_structure = (\n",
+    "        m_pdf_infer_table_structure[0] if len(m_pdf_infer_table_structure) else \"false\"\n",
+    "    ).lower()\n",
+    "    if strategy == \"hi_res\" and pdf_infer_table_structure == \"true\":\n",
```
shreyanid, line 763: Since the new model is being integrated as a strategy soon, will this parameter also need to be set in that case as well? Maybe we can store a set of all the strategies that use the value of this parameter, for future expansion.

scripts/smoketest.py
```diff
@@ -13,10 +13,22 @@
 skip_inference_tests = os.getenv("SKIP_INFERENCE_TESTS", "").lower() in {"true", "yes", "y", "1"}
 
 
-def send_document(filename, content_type, strategy="fast", output_format="application/json"):
+def send_document(
+    filename,
+    content_type,
```
shreyanid, line 18: did the `content_type` input get used?

yuming-long: nope, will remove it

pipeline-notebooks/pipeline-general.ipynb
yuming-long: only hi_res for now

yuming-long:
> I don't see any functionality tests for this new param in `test_general/api/test_app.py` - could you add a test that the table structure data is only included when the parameter is set to true? (similar to the one in your description)

@shreyanid, i can add test in unit test, but it will be the same as the one in the smoke test so i don't  think very necessary to duplicate it. And `hi_res` with 16 pages pdf will take some time to compute again imo

scripts/smoketest.py
shreyanid: seems like the removal got reverted because tests weren't passing. could you either add usage (maybe pass it in to `data` in the request) or write a comment on why it needs to stay?

shreyanid (approved):

PR #142, opened 2023-07-03
Chore: add retry mechanism to fanout request

yuming-long:
### Summary

Set up retry parameters. By default, retry mechanism is enabled, and the fanout request will retry once by default after 1 second of backoff time.


pipeline-notebooks/pipeline-general.ipynb
```diff
@@ -643,17 +645,30 @@
     "    headers = {\n",
     "        \"unstructured-api-key\": request.headers.get(\"unstructured-api-key\")\n",
     "    }\n",
+    "    \n",
+    "    # Retry parameters\n",
+    "    retry_enabled = bool(os.environ.get(\"UNSTRUCTURED_PARALLEL_RETRY_ENABLED\", False))\n",
```
awalker4, line 650: I'd say default should probably be true.

Also you'll need to explicitly check `var == "true"`, otherwise casting a string to a bool will run into this:
```
In [2]: bool("false")
Out[2]: True
```

yuming-long: thanks, i think if we want default to be true, can i just remove this env?

awalker4: Yeah, that makes sense. If we want to disable we can set retries to 0.

awalker4:
Ooh, also, we might want to ignore certain errors we know shouldn't be retryable. Maybe something like:

```
non_retryable_errors = [400, 401]
if status_code != 200:
   if status_code in non_retryable_codes and retry_attempts > 0:
     retry_attempts -= 1
  else:
    throw the error
```

awalker4 (approved):

PR #147, opened 2023-07-10
Chore: throw 400 error if a PDF is password protected

yuming-long:
### Summary
*  import `pycryptodome` to avoid `pypdf.errors.DependencyError: PyCryptodome is required for AES algorithm` from [api](https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#logEventViewer:group=utic-prod-ecs-unstructured-api;stream=ecs-unstructured-api/utic-prod-ecs-unstructured-api/38b8a0f24f674894b464a8133778b9d0;start=2023-07-10T20:37:16.589Z)
* throw 400 if a PDF is password protected

### Test
* `make docker-start-api`
* confirm 400 error with:
```
curl -X 'POST'  \
'http://0.0.0.0:8000/general/v0/general' \
 -F 'files=@sample-docs/layout-parser-paper-password-protected.pdf' \
 | jq -C . | less -R
```


prepline_general/api/general.py
```diff
@@ -68,7 +67,7 @@ def is_expected_response_type(media_type, response_type):
     os.environ["UNSTRUCTURED_ALLOWED_MIMETYPES"] = DEFAULT_MIMETYPES
 
 
-def get_pdf_splits(pdf, split_size=1):
+def get_pdf_splits(pdf_page_splits, split_size=1):
```
ryannikolaidis, line 70: Why the rename pdf->pdf_page_splits here?

ryannikolaidis (commented):
quick clarification on the renaming, otherwise this looks great!

prepline_general/api/general.py
yuming-long: i think i was just trying to avoid read the PDF twice with `pdf = PdfReader(file)` and then pdf is passed down as `pdf_page_splits=pdf.pages`, thats how we use the pdf in the functions, which is a list of pdf splits

awalker4:
Minor nit on the rename - I'd call it `pdf_pages` to clarify that this is before we do the split. Otherwsie LGTM!

PR #148, opened 2023-07-11
Expose model name in api

rbiseck3:
### Description
* Bump `unstructured` dependency to `0.8.1`
* Expose a new parameter in the api to allow a user to pass in an explicit model name for `hi_res`
* Return `403` if they request coordinates but also the new chipper model as that's not possible at the moment
* Update Dockerfile to initialize the chipper mode and use multistage builds


prepline_general/api/general.py
```diff
@@ -190,6 +189,13 @@ def partition_pdf_splits(
     return results
 
 
+logger = logging.getLogger("unstructured_api")
```
yuming-long, line 192: we set up logger in `logger_config.yaml` instead inside codes

prepline_general/api/general.py
```diff
@@ -190,6 +189,13 @@ def partition_pdf_splits(
...
+handler = logging.StreamHandler()
+formatter = logging.Formatter("%(asctime)s %(processName)-10s %(levelname)-8s %(message)s")
+handler.setFormatter(formatter)
+logger.addHandler(handler)
```
yuming-long, line 196: we set up logger in logger_config.yaml instead of inside codes, and those lines have default setting

rbiseck3: What is the appropriate way to map the logger in the code? I'd like to have some logging statements that can associate with the log level (i.e. debug).

yuming-long: you could just remove those lines i think? in `logger_config.yaml`
```
unstructured_api:
    level: DEBUG
    handlers:
      - standard_handler
    propagate: no
```
`standard_handler` already have the formatter setting, and you could also create your own

prepline_general/api/general.py
```diff
@@ -218,6 +237,14 @@ def pipeline_api(
     show_coordinates_str = (m_coordinates[0] if len(m_coordinates) else "false").lower()
     show_coordinates = show_coordinates_str == "true"
 
+    hi_res_model_name = m_hi_res_model_name[0] if len(m_hi_res_model_name) else None
+
+    if hi_res_model_name and hi_res_model_name in CHIPPER_MODEL_TYPES and show_coordinates:
+        raise HTTPException(
+            status_code=403,
```
six5532one, line 244: I think an HTTP 400 would be more suitable than an HTTP 403 because they are requesting an unsupported action rather than an action that specifically their account isn't authorized for.

six5532one (changes requested):

six5532one (approved):

PR #154, opened 2023-07-14
refactor logging from multiline to formatted json

rbiseck3:
### Description
To not bloat our logs, convert the multiline logging of params being passed around to a single line with a well formatted json. This should help with how cloudwatch parses that json if we need to look through it at any point.


tabossert (approved):
LGTM!

PR #165, opened 2023-07-26
Update pip in makefile

shreyanid:
Update pip version to latest (23.2.1)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PIPELINE_FAMILY := general
 PIPELINE_PACKAGE := general
 PACKAGE_NAME := prepline_${PIPELINE_PACKAGE}
-PIP_VERSION := 23.1.2
+PIP_VERSION := 23.2.1
 ARCH := $(shell uname -m)
 
 .PHONY: help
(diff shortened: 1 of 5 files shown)


cragwolfe:
it would be worth running `make pip-compile` as well to make sure that isn't broken (which implies some minor bumps in requirements/*.txt), bumping the version of the API here https://github.com/Unstructured-IO/unstructured-api/blob/main/preprocessing-pipeline-family.yaml#L2 (i.e. to 0.0.32), but then running `make generate-api` to update the api based on the new version.

cragwolfe:
ok, probably worth pinning unstructured to 0.8.4 in https://github.com/Unstructured-IO/unstructured-api/blob/main/requirements/base.in  and re-running `make pip-compile` since that "date" behaviour may change very soon.

shreyanid:
Ran `make pip-compile` to update requirements, but reverted `unstructured` and `unstructured-inference` back to be the same as main so the date field wouldn't cause the `test_parallel_mode_correct_result()` test to fail. 

If this is not a good idea, I will wait to merge this until the metadata date field changes are merged.

PR #167, opened 2023-07-28
Chore: add image table support

yuming-long:
### Summary
* refactored the parameters so its sort alphabetically for partition kwargs
* refactored how we remove field from return elements, move them before deciding cvs/json output so that we don't need to remove twice, also remove the field `date` since it should be fixed in core image 
* added parameter `skip_infer_table_types` with unit test
* bump unstructured library to latest `0.8.7`

### Test
* run `make  run-web-app`
* curl without `skip_infer_table_types` and you won't see `text_as_html` field on the return elements for the table
```
curl -X 'POST'  'http://0.0.0.0:8000/general/v0/general'  \
 -H 'accept: application/json'   \
-H 'Content-Type: multipart/form-data'  \
-F 'files=@sample-docs/layout-parser-paper-with-table.jpg' \
-F 'strategy=hi_res' \
| jq -C . | less -R
```
* curl with`skip_infer_table_types=["pdf"]` and you will `text_as_html` field on the return elements for the table
```
curl -X 'POST'  'http://0.0.0.0:8000/general/v0/general'  \
 -H 'accept: application/json'   \
-H 'Content-Type: multipart/form-data'  \
-F 'files=@sample-docs/layout-parser-paper-with-table.jpg' \
-F 'strategy=hi_res' \
-F 'skip_infer_table_types=pdf' \
| jq -C . | less -R
```


shreyanid:
Your test worked as described. Let me make sure I understand the expected functionality (mostly based on README description):
- By default, table extraction is skipped for images and pdfs
- You can use the `skip_infer_table_types` parameter to specify file types to disable text extraction for. So for example to not extract pdf tables, include `skip_infer_table_types=pdf`
- In this set of tests, the file being partitioned was an image. The image does not extract tables by default, but does that mean the only way to enable the table extraction is by disabling pdf with `skip_infer_table_types=pdf`? There’s no way to directly enable image table extraction without the mention of other unrelated file types to disable?

test_general/api/test_app.py
```diff
@@ -149,6 +149,37 @@ def test_ocr_languages_param():
     assert elements[3]["text"].startswith("안녕하세요, 저 희 는 YGEAS 그룹")
 
 
+def test_skip_infer_table_types_param():
+    """
+    Verify that we skip table instruction unlesee specified
```
shreyanid, line 154: nit
```suggestion
    Verify that we skip table instruction unless specified
```

README.md
```diff
@@ -111,6 +111,22 @@ To extract the table structure from PDF files using the `hi_res` strategy, ensur
   | jq -C . | less -R
 ```
 
+#### Skip Table Extraction
+
+Currently, we provide support for enabling and disabling table extraction for file types other than PDF files. Set parameter `skip_infer_table_types` to specify the document types that you want to skip table extraction with. By default, we skip table extraction
+for PDFs and Images, which are `pdf`, `jpg` and `png`. Again, please note that table extraction only works with `hi_res` strategy. For example, if you only want to skip table extraction for PDFs, you can run:
```
shreyanid, line 117:
```suggestion
for PDFs and Images, which are `pdf`, `jpg` and `png`. Again, please note that table extraction only works with `hi_res` strategy. For example, if you want to extract tables in images and only skip table extraction for PDFs, you can run:
```
maybe add this for extra clarity?

shreyanid (approved):
Changes look good! See earlier comment about expected functionality - it's fine if that's what is expected, just wanted to confirm that was the case. Suggested a README addition if that helps

yuming-long:
> There’s no way to directly enable image table extraction without the mention of other unrelated file types to disable?

Yes, basically we do that bc we don't want to limit on one doc type (image).
we do have a similar param called `pdf_infer_table_structure` to enable table extraction for pdf. while we could follow the pattern and call the one for image individually as `image_infer_table_structure`, i do believe we want to skip/enable table extraction for more doc types in the future, and this pattern requires us to add one parameter to the core image each time, which are hard to keep track and not user friendly. see pr for param in core library:https://github.com/Unstructured-IO/unstructured/pull/973

README.md
yuming-long: i think i will change it to passing an empty parameter, i feel like skip pdf is unnecessary in this example

PR #168, opened 2023-07-29
chore: no chipper model in docker build

cragwolfe:
Previously, the Dockerfile was attempting to bake in the 2.6GB chipper model, but was no longer baking in the default detectron2 model. This was leading to errors when attempting to build the image in CI due to size issues. In addition, it's actually better to pre-load detectron2 in the first place rather than Chipper since it is the preferred "hi_res" model (as used to be the case).

This PR corrects that by only baking in the detectron2 image.

Since the chipper model is no longer baked into the image, there is the option of warming/downloading the chipper model when the application starts in scripts/app-start.sh. By default, it is not downloaded.*

*If it is not downloaded, it will be loaded on demand the first time it is requested. Generally, the download can take about a minute so that first request would be extra slow.

**Test Instructions**

```

# witness the application starting without the download

make docker-build
make docker-start-api

# verify the 1st and 2nd http request for a standard "hi_res" request are about the same, i.e. that a model download was not triggered

time curl -X POST 'http://localhost:8000/general/v0/general' -H 'accept: application/json' \
 -H 'Content-Type: multipart/form-data'  -F "files=@layout-parser-paper-fast.jpg" \
 -F "strategy=hi_res"

# now stop the docker container and start again pre-warming chipper. you'll see output like:
# warming chipper model
# Downloading (…)lve/main/config.json: 100%|████████ ....

docker run -p 8000:8000 --mount type=bind,source=$(realpath .),target=/home/notebook-user/local -e "UNSTRUCTURED_DOWNLOAD_CHIPPER=true" \
  -t --rm pipeline-family-general-dev:latest scripts/app-start.sh
```

In theory, the following command would cause chipper to be invoked:
```
time curl -X POST 'http://localhost:8000/general/v0/general' -H 'accept: application/json' \
   -H 'Content-Type: multipart/form-data'  -F "files=@layout-parser-paper-fast.jpg" \
     -F "hi_res_model_name=chipper" -F "strategy=hi_res"
```
but that does not seem to be the case and should be addressed in a separate issue + PR if confirmed (i also tested with latest version of unstructured, currently 0.8.7).

Bonus changes:

* move Docker entrypoint to a script to support downloading of the chipper model
* there was a complaint when building the docker image that /usr/local/bin/pip  already existed, so don't attempt to overwrite that anymore
* Switch to
```
    uses: docker/setup-buildx-action@v1
```

when building docker images in CI for consistency with unstructured.


Makefile
```diff
@@ -78,7 +78,7 @@ docker-build:
 
 .PHONY: docker-start-api
 docker-start-api:
-	docker run -p 8000:8000 --mount type=bind,source=$(realpath .),target=/home/notebook-user/local -t --rm pipeline-family-${PIPELINE_FAMILY}-dev:latest --log-config logger_config.yaml --host 0.0.0.0 --port 8000
+	docker run -p 8000:8000 --mount type=bind,source=$(realpath .),target=/home/notebook-user/local -t --rm pipeline-family-${PIPELINE_FAMILY}-dev:latest scripts/app-start.sh
```
awalker4, line 81: I don't know if we had a `-i` that got lost somewhere along the way - but could you add that here? Now I can't ctrl c out of the instance.

awalker4 (approved):

PR #173, opened 2023-08-02
fix: Clear space in CI for the image builds

awalker4:
From testing, this frees up 10+ GB on the root filesystem.


tabossert (approved):
LGTM!

PR #174, opened 2023-08-04
Chore: bump unstructured library to 0.9.0

yuming-long:
### Summary

* bumped library
* Rename `date` field to `last_modified`, which is still dropped from return json


awalker4 (approved):
LGTM!

PR #176, opened 2023-08-04
fix: share the cache from the CI workflow

awalker4:
This isn't necessarily a fix for the missing .venv, but this will remove an extra 2GB cache called ci-venv--{commit} and reuse ci-venv-general-{commit}.

Also, don't delete the $AGENT_TOOLSDIRECTORY dir. Testing shows this is `/opt/hostedtoolcache`. No immediate connection to the cache action, which seems to use `/home/runner/work/_temp`, but I did manage to get past the bad interpreter error by keeping this dir.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -9,7 +9,7 @@ env:
   DOCKER_REPOSITORY: quay.io/unstructured-io/unstructured-api
   DOCKER_BUILD_REPOSITORY: quay.io/unstructured-io/build-unstructured-api
   PACKAGE: "unstructured-api"
-  PIPELINE_PACKAGE: "general"
+  PIPELINE_FAMILY: "general"
   PIP_VERSION: "22.2.1"
   PYTHON_VERSION: "3.8"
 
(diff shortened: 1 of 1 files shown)


ryannikolaidis:
did you push everything up? all I see is a rename of PIPELINE_PACKAGE-> PIPELINE_FAMILY

awalker4:
The issue was that the cache key is `ci-venv-${{ env.PIPELINE_FAMILY }}-${{ hashFiles('requirements/test.txt') }}`. It's set properly in the CI job as `ci-venv-general-...` but then this job makes a s2n cache called `ci-venv--...`

ryannikolaidis:
> The issue was that the cache key is `ci-venv-${{ env.PIPELINE_FAMILY }}-${{ hashFiles('requirements/test.txt') }}`. It's set properly in the CI job as `ci-venv-general-...` but then this job makes a 2nd cache called `ci-venv--...`

Ahhh, okay, makes perfect sense. Was wondering if something like that was the case.

ryannikolaidis (approved):
lgtm!

PR #177, opened 2023-08-04
Update pip in makefile

shreyanid:
Update pip version to latest (23.2.1). Bumped requirements through `make pip-compile` and updated api with `make generate-api`

(Abandoning previous version of this PR: #165 )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.0.35
+## 0.0.35-dev0
 
 * Fix a misleading error in make docker-test
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.0.35-dev0
+## 0.0.35
 
 * Fix a misleading error in make docker-test
 
--- a/prepline_general/api/general.py
+++ b/prepline_general/api/general.py
@@ -481,7 +481,7 @@ def ungz_file(file: UploadFile, gz_uncompressed_content_type=None) -> UploadFile
 
 
 @router.post("/general/v0/general")
-@router.post("/general/v0.0.35/general")
+@router.post("/general/v0.0.35-dev0/general")
 def pipeline_1(
     request: Request,
     gz_uncompressed_content_type: Optional[str] = Form(default=None),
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+## 0.0.35
```
cragwolfe, line 1: might aw well call it `0.0.35-dev0`

prepline_general/api/general.py
```diff
@@ -481,7 +481,7 @@ def return_content_type(filename):
 
 
 @router.post("/general/v0/general")
-@router.post("/general/v0.0.35/general")
+@router.post("/general/v0.0.35-dev0/general")
```
yuming-long, line 484: okay to use `0.0.35` because we only publish non dev version to dev/prod API, so no need to change this here (and `preprocessing-pipeline-family.yaml`. We do keep a dev version in Changelog just to keep track which one we want to publish :)

yuming-long (commented):
Added version comment

CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 0.0.35-dev0
+## 0.0.35
```
yuming-long, line 1: this one only can have dev version, and we bump this to non dev when we are ready to publish 0.0.35

prepline_general/api/general.py
shreyanid: hm ok. I get the structure of when to use `-dev` and releases for `unstructured`, but is that process different for the API? when do we use dev for the api, and as per your comment is it only added to the changelog (but then the version would be out of sync with `preprocessing-pipeline-family.yaml`?

CHANGELOG.md
shreyanid: Oh, but then `check-version` would be out of sync right?

yuming-long: nope, they won't check changlog, check-version check if the version is in sync between the yaml file and the version in general api

prepline_general/api/general.py
yuming-long: yeah changlog is just for our purpose to keep track which one we should deploy to API and which one we can wait

yuming-long (approved):
LGTM!

Issue #182, opened 2023-08-09
Internal server error - Extracting tables from a PDF file


skarampatakis:
Hi, I m using the following request to the API in order to extract some tables from a PDF file:
```
curl --location --request POST 'http://localhost:8000/general/v0/general' \
--form 'strategy="hi_res"' \
--form 'pdf_infer_table_structure="true"' \
--form 'files=@"TelecomArgentina_Report 2020.pdf"' \
--form 'ocr_languages="eng"' \
--form 'skip_infer_table_types=""'
```

The request fails after 14 mins, I see the following on the logs:

```
2023-08-09 06:52:34,237 172.17.0.1:46490 POST /general/v0/general HTTP/1.1 - 500 Internal Server Error
2023-08-09 06:52:34,238 uvicorn.error ERROR Exception in ASGI application
Traceback (most recent call last):
  File "/home/notebook-user/.local/lib/python3.8/site-packages/uvicorn/protocols/http/httptools_impl.py", line 426, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/notebook-user/.local/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 84, in __call__
    return await self.app(scope, receive, send)
  File "/home/notebook-user/.local/lib/python3.8/site-packages/fastapi/applications.py", line 289, in __call__
    await super().__call__(scope, receive, send)
  File "/home/notebook-user/.local/lib/python3.8/site-packages/starlette/applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/notebook-user/.local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/home/notebook-user/.local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/home/notebook-user/.local/lib/python3.8/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/home/notebook-user/.local/lib/python3.8/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "/home/notebook-user/.local/lib/python3.8/site-packages/fastapi/middleware/asyncexitstack.py", line 20, in __call__
    raise e
[...]

PR #183, opened 2023-08-11
build(release): bump unstructured

shreyanid:
Related to downstream issue:

https://github.com/Unstructured-IO/unstructured-api/issues/182

And upstream PR:

https://github.com/Unstructured-IO/unstructured-inference/pull/165


pipeline-notebooks/pipeline-general.ipynb
```diff
@@ -921,6 +921,7 @@
        "[{'type': 'UncategorizedText',\n",
        "  'element_id': 'db1ca22813f01feda8759ff04a844e56',\n",
        "  'metadata': {'filename': 'family-day.eml',\n",
+       "   'file_directory': '/Users/shreyanid/Documents/all-unstructured/unstructured-api/sample-docs',\n",
```
awalker4, line 924: Looks like this field has been in the metadata for a while - did something just change for it to show here? This will probably get us into trouble with `make check-notebooks`.

shreyanid: That's exactly what's happening :') Not sure what just changed, [this commit](https://github.com/Unstructured-IO/unstructured/commit/f8a7ae2953ff4dfb06cc42ca9fc62a36a053fab0) was thrown out as a possibility. But it isn't happening on the current main branch of api, so that means some package bump (likely unstructured) caused the change. Would appreciate some help debugging where the change came from, and why the difference in file directory is not happening locally but happening in CI.

shreyanid: In [this](https://github.com/Unstructured-IO/unstructured/compare/0.9.1...0.9.2) release comparison it doesn't seem like anything recently changed related to the file_directory field

awalker4: Hmm, that's a fun one! I have some cycles now to take a look as well.

awalker4: Ah, I think it is that commit. [Here](https://github.com/Unstructured-IO/unstructured/blob/79a1eb86837ccfd84eb265985d6981d9ef9fe965/unstructured/documents/elements.py#L173) it looks like `file_directory` will be set whenever a filename is present, and that commit will have us sending `metadata_filename` all the time. All the different filename params confuse me, but tldr is that we should remove that field like the ones [over here](https://github.com/Unstructured-IO/unstructured-api/blob/c9b74d43e716fb36eb2c07b62ae3eb237feb4e15/prepline_general/api/general.py#L367).

awalker4 (approved):
LGTM! I'm optimistically assuming the notebook check passed this time 😄

PR #186, opened 2023-08-17
fix: fix "not a valid pdf error" in parallel mode

awalker4:
This was a subtle bug that came out in the retry logic. When we get a 500 during the requests.post, we'll try again. However, the pdf was stored in a BytesIO, which had already been read the first time we sent it. The next request sends an empty file, which results in a 400 response masking the original error.

Steps to verify:

* First, checkout `main`
* Start up the api in parallel mode

```
make run-web-app
```

* Insert a 500 error into `prepline_general/api/general.py:partition_pdf_splits()`
```
        raise HTTPException(status_code=500)  # Throw an error here
        return partition(
            file=file, file_filename=file_filename, content_type=content_type, **partition_kwargs
        )

```

* Send a document and see that the 500 is hidden behind a 400 error
```
$ curl 'http://localhost:8000/general/v0/general' --header 'Accept: application/json' --form files=@sample-docs/layout-parser-paper-fast.pdf

{"detail":"layout-parser-paper-fast.pdf does not appear to be a valid PDF"}%
```

* Switch to this branch and do it again - you should now get a 500 `Internal server error` response

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.0.36-dev0
+
+* Fix a bug in parallel mode causing `not a valid pdf` errors
+
 ## 0.0.35
 
 * Bump unstructured library to 0.9.2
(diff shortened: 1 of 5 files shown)


cragwolfe (approved):
:clap::clap:

Issue #189, opened 2023-08-18
bug/parallel mode: unsupported operand types


awalker4:
![Screenshot 2023-08-18 at 1 44 04 PM](https://github.com/Unstructured-IO/unstructured-api/assets/3846783/05385b8b-127c-4130-b33f-693aabd1d1ad)

I don't have the file that caused this, but `element.metadata.page_number` is apparently `None`.

yuming-long:
found this error while i was testing dev with some code from a user:
```
import aiohttp
from aiohttp import FormData
url = 'https://dev.api.unstructured.io/general/v0/general'
file_path = 'example-docs/layout-parser-paper-fast.pdf'
async def send_request():
    async with aiohttp.ClientSession() as session:
        with open(file_path, 'rb') as file:
            file_content = file.read()
            data = FormData()
            # Add the file field
            data.add_field('files', file_content, filename=file_path, content_type='application/octet-stream')
            # Add other fields
            data.add_field('ocr_languages', "eng")
            data.add_field('strategy', "ocr_only" if file_path.lower().endswith((".jpeg", ".jpg", ".png")) else "auto")
            data.add_field('include_page_breaks', "true")
            data.add_field('pdf_infer_table_structure', "true")
            headers = {
                'unstructured-api-key': <api-key>
            }
            async with session.post(url, data=data, headers=headers) as response:
                response_text = await response.text()
                print(response_text)
```

PR #196, opened 2023-08-22
Chore: fix bug caused by page break has no page number

yuming-long:
close: https://github.com/Unstructured-IO/unstructured-api/issues/189

we are having this bug because page number is None if we include page breaks.


awalker4 (approved):
Awesome!

PR #199, opened 2023-08-23
fix: fix occasional UnboundLocalError

awalker4:
To verify, you need to run parallel mode and send a pdf file with the correct mimetype but a wrong file extension.

On main, start the server in parallel mode and confirm the bug happens:
```
make run-web-app
```

In a python shell:
```
import requests

with open("sample-docs/layout-parser-paper.pdf", "rb") as f:
    res = requests.post("http://localhost:8000/general/v0/general",
                files={"files": ("foo.txt", f, "application/pdf")})

    # Should be Internal server error
    print(res.text)

```

Then try again on this branch.

Also, removed logging of the request object, which just displays an object reference next to the params.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
-## 0.0.38-dev0
+## 0.0.38-dev1
 
 * Fix page break has None page number bug
+* Fix a UnboundLocalError using pdfs in parallel mode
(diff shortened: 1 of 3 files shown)


yuming-long (approved):
LGTM after test!

PR #200, opened 2023-08-23
fix: param skip_infer_table_types should be first item from m_skip_infer_table_types 

yuming-long:
forgot `m_skip_infer_table_types` is a list of str


awalker4 (approved):
LGTM! Is this reflected in any relevant readme examples?

yuming-long:
readme and github doc can keep the same :)

PR #215, opened 2023-09-01
fix: fix parameter renaming bug when using parallel mode

awalker4:
This affected `hi_res_model_name` - the api param is different than the partition param so it wasn't passed correctly. Add a unit test to catch all future cases of this. If the param is passed correctly, you can send a bogus model name and see it get rejected by unstructured-inference.

Run the following on main - you'll get back a full response. Then try it on this branch - the invalid model name is passed through and you get back a 500. (Separate issue to turn this into a 400...)
```
make run-web-app

# This should be an error
curl 'http://localhost:8000/general/v0/general' --form files=@sample-docs/layout-parser-paper-fast.pdf --form strategy=hi_res --form hi_res_model_name=yoloxdd
```


prepline_general/api/general.py
```diff
@@ -169,7 +169,12 @@ def partition_pdf_splits(
     pages_per_pdf = int(os.environ.get("UNSTRUCTURED_PARALLEL_MODE_SPLIT_SIZE", 1))
 
     # If it's small enough, just process locally
+    # (Some kwargs need to be renamed for local partition)
     if len(pdf_pages) <= pages_per_pdf:
+        if "hi_res_model_name" in partition_kwargs:
+            partition_kwargs["model_name"] = partition_kwargs["hi_res_model_name"]
```
yuming-long, line 175: do we want to rename `hi_res_model_name` to `model_name` for consistency? (why we have two different names two

awalker4: That's a good question - might have been to make it more clear when the param was relevant. I think we'll want to revisit all the params soon. We've maybe outgrown the api-tools pattern of making everything a list. We'll need to keep the old name for backwards compat, but this would be a good chance to make those consistency fixes.

prepline_general/api/general.py
```diff
@@ -169,7 +169,12 @@ def partition_pdf_splits(
...
+            del partition_kwargs["hi_res_model_name"]
```
badGarnet, line 176:
```suggestion
            partition_kwargs["model_name"] = partition_kwargs.pop("hi_res_model_name")
```

jasonbot (approved):

PR #216, opened 2023-09-01
feat: add exponential retries to parallel mode calls

awalker4:
Replace our parallel mode retry logic with the `backoff` library. This gives us exponential backoff, retryable error codes, etc with just a decorator, which really cleans up the code.

Changes:
* Refactor `partition_file_via_api` and move the request with backoff to `call_api`
* Add `backoff` as a dependency and `pip compile`
* Make sure we don't dump api parameters on every parallel call
* Don't allow internal calls to bypass the 503 low memory gate (Should be handle in the retries like everything else)

To test this, try adding an HTTPException to the code.

Add a non-retryable exception in `partition_pdf_splits`:
```
    # (Some kwargs need to be renamed for local partition)
        raise HTTPException(status_code=400)

```

When you run this and send a file, you'll get the 400 back immediately:
```
export UNSTRUCTURED_PARALLEL_NUM_THREADS=1

make run-web-app


curl -X POST 'http://localhost:8000/general/v0/general' --form files=@sample-docs/layout-parser-paper.pdf
{"detail":"Bad Request"}
```

Now, return a 500 error instead and run again. In this case you'll get a server error, but in the logs you should see that the retries happened:
```
Giving up call_api(...) after 3 tries (fastapi.exceptions.HTTPException)
```


yuming-long:
Tested the 400 not retry able works!

For the 500 error code, i made one api request to test it - from my terminal i see 13 `POST /general/v0/general HTTP/1.1 - 500 Internal Server Error` log and four `Giving up call_api(...) after 3 tries (fastapi.exceptions.HTTPException)` 
i don't think this is the right case, should we only see one `POST /general/v0/general HTTP/1.1 - 500 Internal Server Error` from one external request and maybe one `Giving up call_api(...) after 3 tries (fastapi.exceptions.HTTPException)`
also i think we should mock this in unit tests (add more checks to log info?)

awalker4:
Aha! Figured out what happened. The test steps were supposed to say `export UNSTRUCTURED_PARALLEL_MODE_THREADS=1`. Otherwise we'll default to 3 parallel calls that are all retrying at the same time. I updated the "don't retry" test to use a mock so we can assert the number of calls.

awalker4:
@yuming-long let me know how it looks with that change!

test_general/api/test_app.py
```diff
@@ -549,34 +530,36 @@ def mock_response(*args, **kwargs):
...
-        data={"pdf_processing_mode": "parallel"},
     )
 
     assert response.status_code == 401
-    assert mock_sleep.call_count == 0
+
+    # Often page 2 will start processing before the page 1 exception is raised.
+    # So we can't assert called_once, but we can assert the count is less than it
+    # would have been if we used all retries.
+    assert remote_partition.call_count < 4
```
yuming-long, line 562: why this needs retry tho? should the 401 error from page 1 be thrown immediately since its not retryable?

awalker4: Yeah, asserting on `called_once` should work like that - we call the mock, raise the error, and bail. The problem I'm having is with the ThreadpoolExecutor.map that we use. The first page raises an exception, but then before we retrieve it in the results, the second page goes through and also hits the error. When I run this locally, the call usually happens twice. Instead of accounting for timing of the threads, we can just assert that call_count is below 4 (3 retries + the original).

yuming-long (approved):
LGTM! 🎉

PR #222, opened 2023-09-06
chore: Cut ties with unstructured-api-tools

awalker4:
We've moved in a different direction and the api-tools workflow doesn't make sense. Instead of modifying a notebook and running `make generate-api`, let's just work with the code directly from here on out.

* Remove the `pipeline_notebook` dir and preprocessing yaml file
* Remove the notebook consistency scripts
* Update `make version-sync` to bump the version in `general.py` directly
* Remove all notebook centric makefile targets


preprocessing-pipeline-family.yaml
```diff
@@ -1,2 +0,0 @@
-name: general
```
yuming-long, line 1: looks like this file is used in docker build ci: https://github.com/Unstructured-IO/unstructured-api/blob/main/.github/workflows/docker-publish.yml#L140
also i feel like we could still keep this file?

yuming-long (approved):
LGTM! (tho not sure why api_consistency is still in ci)

PR #249, opened 2023-09-24
fix: Do not return 400 error for encrypted but readable pdfs

### LINKED ISSUE #236: Receive unexpected status code 400 from the API.
I am running this in Python and my first PDF worked fine with the API. The first PDF was 38 pages.

I am trying now with a PDF which is 98 pages and I receive the error: "Receive unexpected status code 400 from the API.".

Any idea why?

awalker4:
Fixes #236 

The `pdf.is_encrypted` check is true for files with edit protections, so we were returning 400 "This file is encrypted" for files that were perfectly readable. To verify, run this branch with `make run-web-app` and submit the pdf in the linked issue.

Other changes:
* Remove filename from some error messages. This can get way too verbose.

### DIFF
--- a/test_general/api/test_app.py
+++ b/test_general/api/test_app.py
@@ -656,31 +656,31 @@ def test_encrypted_pdf():
     client = TestClient(app)
     test_file = Path("sample-docs") / "layout-parser-paper-fast.pdf"
     original_pdf = PdfReader(test_file)
-    temp_file = tempfile.NamedTemporaryFile()
 
-    # This file is user encrypted and cannot be read
-    writer = PdfWriter()
-    writer.append_pages_from_reader(original_pdf)
-    writer.encrypt(user_password="password123")
-    writer.write(temp_file.name)
+    with tempfile.NamedTemporaryFile() as temp_file:
+        # This file is user encrypted and cannot be read
+        writer = PdfWriter()
+        writer.append_pages_from_reader(original_pdf)
+        writer.encrypt(user_password="password123")
+        writer.write(temp_file.name)
 
-    # Response should be 400
-    response = client.post(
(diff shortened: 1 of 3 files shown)


test_general/api/test_app.py
```diff
@@ -576,25 +575,6 @@ def test_partition_file_via_api_not_retryable_error_code(monkeypatch, mocker):
     assert remote_partition.call_count < 4
 
 
-def test_password_protected_pdf():
-    """
-    Verify we get a 400 error if the PDF is password protected
-    """
-    client = TestClient(app)
-    # a password protected pdf file, password is "password"
-    test_file = Path("sample-docs") / "layout-parser-paper-password-protected.pdf"
```
yuming-long, line 585: please remove this file as well

yuming-long (approved):
LGTM!

test_general/api/test_app.py
awalker4: ooh, good call!

test_general/api/test_app.py
```diff
@@ -666,3 +646,41 @@ def test_chunking_strategy_additional_params():
...
+
+def test_encrypted_pdf():
+    """
+    Test that we throw an error if a pdf is password protected.
+    A pdf can be encrypted but still readable - don't throw an error here.
+    """
+    client = TestClient(app)
+    test_file = Path("sample-docs") / "layout-parser-paper-fast.pdf"
+    original_pdf = PdfReader(test_file)
+    temp_file = tempfile.NamedTemporaryFile()
```
yuming-long, line 659: some nit: we should use temp file with `with tempfile.NamedTemporaryFile() as temp_file:` so that we close the temp file

Issue #237, opened 2023-09-16
TypeError: __init__() got an unexpected keyword argument 'detection_class_prob'


Jimchoo91:
I tried uploading a 50 page PDF using the API, and received this error?

TypeError: __init__() got an unexpected keyword argument 'detection_class_prob'

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-199-16b05c99d0ba> in <module>
     10 )
     11 
---> 12 docs = loader.load()

/opt/anaconda3/lib/python3.8/site-packages/langchain/document_loaders/unstructured.py in load(self)
     84     def load(self) -> List[Document]:
     85         """Load file."""
---> 86         elements = self._get_elements()
     87         if self.mode == "elements":
     88             docs: List[Document] = list()

/opt/anaconda3/lib/python3.8/site-packages/langchain/document_loaders/unstructured.py in _get_elements(self)
    267 
    268     def _get_elements(self) -> List:
--> 269         return get_elements_from_api(
    270             file_path=self.file_path,
    271             api_key=self.api_key,

/opt/anaconda3/lib/python3.8/site-packages/langchain/document_loaders/unstructured.py in get_elements_from_api(file_path, file, api_url, api_key, **unstructured_kwargs)
    202         from unstructured.partition.api import partition_via_api
    203 
--> 204         return partition_via_api(
    205             filename=file_path,
    206             file=file,

/opt/anaconda3/lib/python3.8/site-packages/unstructured/partition/api.py in partition_via_api(filename, content_type, file, file_filename, api_url, api_key, **request_kwargs)
     86                 "metadata_filename must be specified as well.",
     87             )
---> 88         files = [
     89             ("files", (metadata_filename, file, content_type)),  # type: ignore
     90         ]

/opt/anaconda3/lib/python3.8/site-packages/unstructured/staging/base.py in elements_from_json(filename, text, encoding)
    117     """Loads a list of elements from a JSON file or a string."""
    118     exactly_one(filename=filename, text=text)
--> 119 
    120     if filename:
    121         with open(filename, encoding=encoding) as f:

/opt/anaconda3/lib/python3.8/site-packages/unstructured/staging/base.py in dict_to_elements(element_dict)
    100                     metadata=metadata,
    101                 ),
--> 102             )
    103 
    104     return elements

/opt/anaconda3/lib/python3.8/site-packages/unstructured/staging/base.py in isd_to_elements(isd)
[...]

PR #252, opened 2023-09-25
fix: remove `parent_id` until backwards compat bug is addressed

awalker4:
Original issue: https://github.com/Unstructured-IO/unstructured-api/issues/237
Core library fix: https://github.com/Unstructured-IO/unstructured/pull/1526

Anyone who calls `partition_via_api` will hit this bug until they upgrade `unstructured`, which includes any Langchain users of `UnstructuredAPIFileLoader`. The immediate fix is to remove `parent_id` from the hosted api. Next, we can ensure that [langchain users](https://github.com/langchain-ai/langchain/pull/11025) are up to date. Finally, the core library fix above will address any new fields going forward. It will be safe to readd the `parent_id` once users are generally on `unstructured>=0.10.15`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
-## 0.0.47-dev0
+## 0.0.47
 
 * **Adds `chunking_strategy` kwarg and associated params** These params allow users to "chunk" elements into larger or smaller `CompositeElement`s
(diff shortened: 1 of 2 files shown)


awalker4:
Note we should merge https://github.com/Unstructured-IO/unstructured-api/pull/249 before cutting the release as this is another high priority fix.

yuming-long (approved):
LGTM!

PR #253, opened 2023-09-25
chore: run make tidy on api files

awalker4:
These were excluded from linting when they were autogenerated. We need to tidy these files, otherwise we'll get blocked by our internal deploy pipeline.

Note that this should replace the current 0.0.47 candidate.


amanda103 (approved):
lgtm pending passing builds :)

PR #258, opened 2023-09-27
chore: don't let dependabot pull unstructured-inference

awalker4:
Given this fix: https://github.com/Unstructured-IO/unstructured/pull/1474

We should get the correct inference whenever unstructured is bumped. We can't take these out of order. Also tweak the gh action to remove `[local-inference]` when `bumping unstructured` message is added.


yuming-long (approved):
lgtm!

Issue #261, opened 2023-09-29
Docker entry point error


glorat:
https://github.com/Unstructured-IO/unstructured-api/blob/c31465029047fcfc67e401e5b8b75fae95493400/scripts/app-start.sh#L1C17-L1C17

In certain docker environments (e.g. google cloud run) this will fail with
```
env: 'bash ': No such file or directory
env: use -[v]S to pass options in shebang lines
Container called exit(127).
```
Note the space after bash

alex-atkins:
I'm getting the same issue in 0.0.80 when running app-start.sh, but I don't see a white space.  Anyone know if there's another cause?

```
env: can't execute 'bash
': No such file or directory
```

PR #262, opened 2023-09-29
fix: remove spurious whitespace in app-start.sh

glorat:
Resolves issue #261

### DIFF
--- a/prepline_general/api/general.py
+++ b/prepline_general/api/general.py
@@ -569,7 +569,7 @@ def ungz_file(file: UploadFile, gz_uncompressed_content_type=None) -> UploadFile
 
 
 @router.post("/general/v0/general")
-@router.post(f"/general/v{app.version}/general")
+@router.post(f"/general/v0.0.49/general")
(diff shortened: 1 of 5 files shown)


awalker4 (approved):

awalker4:
Can you also bump the changelog and run `make version-sync`?

prepline_general/api/general.py
```diff
@@ -569,7 +569,7 @@ def return_content_type(filename):
 
 
 @router.post("/general/v0/general")
-@router.post(f"/general/v{app.version}/general")
+@router.post(f"/general/v0.0.49/general")
```
cragwolfe, line 572: in retrospect, did this need to change to be hardcoded @awalker4 ?

awalker4: I had to revert this from https://github.com/Unstructured-IO/unstructured-api/pull/256. Turns out general.py is initializing its own `app` and so this causes the endpoint to be `/general/v1.0.0/general`

PR #273, opened 2023-10-05
fix: handle PdfStreamError

### LINKED ISSUE #266: PdfStreamError: Stream has ended unexpectedly
API users are hitting this error on certain files. 

```
PdfStreamError: Stream has ended unexpectedly
  File "prepline_general/api/general.py", line 686, in pipeline_1
    list(response_generator(is_multipart=False))[0] if len(files) == 1 else join_responses(list(response_generator(is_multipart=False)))
  File "prepline_general/api/general.py", line 607, in response_generator
    response = pipeline_api(
  File "prepline_general/api/general.py", line 278, in pipeline_api
    pdf = PdfReader(file)
  File "pypdf/_reader.py", line 332, in __init__
    self.read(stream)
  File "pypdf/_reader.py", line 1554, in read
    self._find_eof_marker(stream)
  File "pypdf/_reader.py", line 1625, in _find_eof_marker
    line = read_previous_line(stream)
  File "pypdf/_utils.py", line 268, in read_previous_line
    raise PdfStreamError(STREAM_TRUNCATED_PREMATURELY)
```

awalker4:
Fixes #266 

We see this error when a user sends a non pdf file, and the content type is set to pdf. We try to load it in `PdfReader`, and see warnings like this:

```
WARNING:pypdf._reader:invalid pdf header: b'\xff\xd8\xff\xe1\x9b'
WARNING:pypdf._reader:EOF marker not found
```

We just need to catch this error and return a useful 400 message. Also, add a general catch to log any other pdf read errors here - we've hit a few others.

To verify, send any non pdf file with content type as pdf:
```python
import requests

filename = "/path/to/jpeg/file"

res = requests.post(
    "http://localhost:8000/general/v0/general",
    files={"files": (filename, open(filename, "rb"), "application/pdf")},
    )

print(res.text)
```

Other changes:
* Do not remove `uvicorn.error` logs locally, these can be useful

### DIFF
--- a/prepline_general/api/general.py
+++ b/prepline_general/api/general.py
@@ -278,13 +278,16 @@ def pipeline_api(
 
             # This will raise if the file is encrypted
             pdf.metadata
-        except pypdf.errors.EmptyFileError:
+        except (pypdf.errors.EmptyFileError, pypdf.errors.PdfStreamError):
             raise HTTPException(status_code=400, detail="File does not appear to be a valid PDF")
         except pypdf.errors.FileNotDecryptedError:
             raise HTTPException(
                 status_code=400,
                 detail="File is encrypted. Please decrypt it with password.",
             )
+        except Exception as e:
+            logger.warn(f"PDF read error: {e}")
+            raise HTTPException(status_code=400, detail="File does not appear to be a valid PDF")
 
     strategy = (m_strategy[0] if len(m_strategy) else "auto").lower()
     strategies = ["fast", "hi_res", "auto", "ocr_only"]
--- a/prepline_general/api/app.py
+++ b/prepline_general/api/app.py
@@ -17,9 +17,10 @@ app = FastAPI(
 )
 
 # Note(austin) - This logger just dumps exceptions
-# We'd rather handle those below
+# We'd rather handle those below, so disable this in deployments
 uvicorn_logger = logging.getLogger("uvicorn.error")
-uvicorn_logger.disabled = True
+if os.environ.get("ENV") in ["dev", "prod"]:
+    uvicorn_logger.disabled = True
 
 
 # Catch all HTTPException for uniform logging and response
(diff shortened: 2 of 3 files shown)


prepline_general/api/general.py
```diff
@@ -278,13 +278,16 @@ def pipeline_api(
...
+        except (pypdf.errors.EmptyFileError, pypdf.errors.PdfStreamError):
             raise HTTPException(status_code=400, detail="File does not appear to be a valid PDF")
         except pypdf.errors.FileNotDecryptedError:
             raise HTTPException(
                 status_code=400,
                 detail="File is encrypted. Please decrypt it with password.",
             )
+        except Exception as e:
+            logger.warn(f"PDF read error: {e}")
+            raise HTTPException(status_code=400, detail="File does not appear to be a valid PDF")
```
yuming-long, line 290: can we combine the two `HTTPException(status_code=400, detail="File does not appear to be a valid PDF")`? (another one on line 282)

awalker4: I added the generic catch to log whatever we just caught. This will let us track anything else that's getting thrown, and decide how to handle it.

prepline_general/api/general.py
```diff
@@ -278,13 +278,16 @@ def pipeline_api(
...
             pdf.metadata
-        except pypdf.errors.EmptyFileError:
+        except (pypdf.errors.EmptyFileError, pypdf.errors.PdfStreamError):
             raise HTTPException(status_code=400, detail="File does not appear to be a valid PDF")
         except pypdf.errors.FileNotDecryptedError:
             raise HTTPException(
                 status_code=400,
                 detail="File is encrypted. Please decrypt it with password.",
             )
+        except Exception as e:
```
cragwolfe, line 288: this is too broad. other things could have gone wrong

awalker4: Fair point, I figured we'd log the additional errors, but they'll be easier to spot as a new 500. Sentry also found some `PdfReadError`, so I'll add another case above.

prepline_general/api/general.py
awalker4: Adjusted this per Crag's note below

cragwolfe (approved):

PR #274, opened 2023-10-06
chore: use generator in parallel mode, other misc fixes

awalker4:
* Yield pdf splits in parallel mode to save memory
* Due to this, a parallel retry test was failing. A multi page pdf will process other pages while retrying the first one, and we can't assert that our function was only hit once. The fix is to use a file with fewer pages. I copied one from unstructured. 
* Add a default to the minimum memory 503 check
* Update pyenv instructions in readme
* Add linebreaks to chunking params in readme


yuming-long:
btw paddle env is not need (won't be an issue if paddle is not installed at all)

awalker4:
@yuming-long mind taking another look at this small pr?

test_general/api/test_app.py
```diff
@@ -617,10 +617,7 @@ def test_partition_file_via_api_not_retryable_error_code(monkeypatch, mocker):
 
     assert response.status_code == 401
 
-    # Often page 2 will start processing before the page 1 exception is raised.
-    # So we can't assert called_once, but we can assert the count is less than it
-    # would have been if we used all retries.
-    assert remote_partition.call_count < 4
+    assert remote_partition.called_once()
```
yuming-long, line 620: may i ask if a multiple page pdf will still fail on this `called_once`

yuming-long (approved):
LGTM!

test_general/api/test_app.py
awalker4: I fixed this by using a pdf with fewer pages! The generator makes parallel mode even faster, so the test kept failing because too many pages get processed.

PR #281, opened 2023-10-12
catch json ValueError and provide example schema

### LINKED ISSUE #271: Detected a JSON file that does not conform to the Unstructured schema. partition_json currently only processes serialized Unstructured output.
This is a ValueError raised from partition(). We need to catch this and transform it to a 400 error. See https://github.com/Unstructured-IO/unstructured/issues/547 for more context.

Coniferish:
Closes #271 

Catches ValueError raised by partition() and transforms it to a 400 error. Also provides an example of the Unstructured schema.


awalker4:
LGTM! I think we can put the example schema in the docs somewhere and keep the error message nice and short.

awalker4 (approved):

Coniferish:
Sounds good. Where in the docs should we add the example, though? Ryan mentioned that there's a [PR ](https://github.com/Unstructured-IO/unstructured/pull/1468) in progress to add it to the docs, but idk when it'll get added

awalker4:
Addressed in #290

PR #290, opened 2023-10-20
fix: fix local variable 'elements' referenced before assignment

awalker4:
We've seen this error in the hosted api:

```
UnboundLocalError: local variable 'elements' referenced before assignment
File "prepline_general/api/general.py", line 649, in response_generator
    response = pipeline_api(
  File "prepline_general/api/general.py", line 464, in pipeline_api
    for i, element in enumerate(elements):
```

The if block in the check for `zipfile.BadZipFile` allows us to fall through without assigning `elements`. Instead, we should raise a 400 error when we catch this exception.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
-## 0.0.54-dev1
+## 0.0.54-dev2
 
 * Bump unstructured to 0.10.24
 * Use a generator when splitting pdfs in parallel mode
 * Add a default memory minimum for 503 check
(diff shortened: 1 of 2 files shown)


cragwolfe (approved):

Coniferish:
LGTM
My only concern is casting too large a net for MS errors and overlooking something else happening. But it makes sense to remove the string check since that seems to be the only message that's returned by `zipfile.BadZipFile`

Coniferish (approved):
LGTM

PR #294, opened 2023-10-25
chore: reintroduce parent_id metadata field

awalker4:
This was removed in #252 to address a backwards compat bug in Langchain. The bug was fixed in unstructured 0.10.15 and this is now the min version in Langchain.


yuming-long (approved):
🎉 LGTM

PR #296, opened 2023-10-26
feat: Only allow one Chipper call at a time

awalker4:
Chipper V2 is very memory hungry. While we work to optimize this, we need to restrict the server to one call at a time. While the model is in use, we'll return a 503 "Please try again". Our hosted API should scale up to meet demand, so the next call should route to an available server.

This includes a refactor to how partition_kwargs are passed to either parallel mode, local partition, or local partition with the new Chipper protection.

To verify, try calling Chipper twice:
```
curl -X POST 'http://localhost:8000/general/v0/general' --form files="@$file" --form strategy=hi_res --form hi_res_model_name=chipper &
curl -X POST 'http://localhost:8000/general/v0/general' --form files="@$file" --form strategy=hi_res --form hi_res_model_name=chipper
```

The second call will get a 503 response.

Other changes:
* Return a 400 error if Chipper isn't loaded. The model is private, make sure we explain this for users who self host
* Pass the huggingface token to `make docker-start-api` for better dev experience
* Add a `make docker-start-bash` while we're in here

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
-## 0.0.55-dev1
+## 0.0.55
 
-* Bump unstructured to 0.0.26
+* Bump unstructured to 0.10.26
 * Bring parent_id metadata field back after fixing a backwards compatibility bug
+* Restrict Chipper usage to one at a time. The model is very resource intense, and this will prevent issues while we improve it.
 
 ## 0.0.54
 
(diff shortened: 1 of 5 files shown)


ryannikolaidis (approved):
running locally in a container on a mac m2 (with docker resources maxed out), I'm hitting a segmentation fault while processing with chipper, but that's outside the concern of this PR. I am able to see with two concurrent requests that the first one has processing attempted while the other immediately gets a 503, which is good enough for me!

PR #300, opened 2023-11-03
chore: remap `chipper` to use `chipperv2`

awalker4:
We want this to alias to the current model. Also, wrap the error that comes from chipperv1 becoming a private model.

To verify:
* Run the server
```
make run-web-app
```
* A call to chipperv1 should give a useful error:
```
curl -X POST 'http://localhost:8000/general/v0/general' --form files="@$file" --form strategy=hi_res --form hi_res_model_name=chipperv1
{"detail":"The Chipper model is not available for download. It can be accessed via the official hosted API."}
```
* Change the model name to `chipper`. You should see `chipperv2` in the log message containing partition input data

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
-## 0.0.56-dev1
+## 0.0.56
 * **Add `max_characters` param for chunking** This param gives users additional control to "chunk" elements into larger or smaller `CompositeElement`s
 * Bump unstructured to 0.10.28
+* Make sure chipperv2 is called whien `hi_res_model_name==chipper`
 
 
 ## 0.0.55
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## 0.0.56-dev1
+## 0.0.56
```
yuming-long, line 1: do we make a new release for this in API?

yuming-long (approved):
* `'hi_res_model_name=chipper'` give me error: "chipper model is not available for download. It can be accessed via the official hosted API." (error from chipper v2)
* `'hi_res_model_name=chipperv1'` gives me the json output (chipper v1 should still work)

yuming-long (commented):
Tested this again after setting the HF token (so previous comment is from without token set)
* output from `chipper v2`: the first element has text: `LayoutParser: A Unified Toolkit for Deep Learning Based Document Image Analysis`
* output from `chipper v1`: the first element has text: `arXiv: 2103.15348v2 [cs.CW] 21 Jun 2021`
* output from `chipper`:  the first element has text: `LayoutParser: A Unified Toolkit for Deep Learning Based Document Image Analysis`, same as `chipperv2`

PR #302, opened 2023-11-07
chore: default to true for multipage_sections

awalker4:
The readme had said the default is true, and this matches the library default.

To verify, make a call to the server and observe the debug log with all the parameters:

```
curl --location 'localhost:8000/general/v0/general' \
--header 'Accept: application/json' \
--form 'files=@"sample-docs/layout-parser-paper.pdf"'
```

In the param dump, you should see: `"multipage_sections": true,`


LaverdeS (approved):
LGTM!

PR #309, opened 2023-11-15
Chore: Bump unstructured to 0.10.30

yuming-long:
^^^
Also releasing this version since `0.10.30` include some 500 error bug fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.0.58
+
+* Bump unstructured to 0.10.30
+
 ## 0.0.57
 * Make sure `multipage_sections` param defaults to `true` as per the readme
 * Bump unstructured to 0.10.29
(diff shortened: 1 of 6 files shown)


awalker4 (approved):
Lgtm!

scripts/hi_res_model_initialize.py
```diff
@@ -0,0 +1,16 @@
...
+    """Download default model or model specified by UNSTRUCTURED_HI_RES_MODEL_NAME environment
+    variable (avoids subprocesses all doing the same)"""
+
+    # If more than one model will be supported and left up to user selection
+    supported_model = os.environ.get("UNSTRUCTURED_HI_RES_SUPPORTED_MODEL", "")
+    if supported_model:
+        for model_name in supported_model.split(","):
+            get_model(model_name=model_name)
+
+    get_model(os.environ.get("UNSTRUCTURED_HI_RES_MODEL_NAME"))
```
cragwolfe, line 16: This is fine for the quick workaround, but it seems like this initialization code should be moved within unstructured itself, out of the ingest module, maybe into its own top-level module.

yuming-long: yeah i was thinking maybe under `partition` dir?

cragwolfe: that works

PR #311, opened 2023-11-16
chore: Remove huggingface token passthrough in makefile

awalker4:
This variable isn't relevant when developing with the public image.


yuming-long (approved):
LGTM!

PR #316, opened 2023-11-25
catch 500 error for fast strategy

### LINKED ISSUE #314: catch 500 error and return to users when using `fast` strategy for images
**Describe the bug**
We don't support ay fallback for `fast` strategy on images anymore, and unst will raise [value error](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/strategies.py#L22) if partition uses `fast` strategy for image files. We should catch this 500 error, raise a `400 HTTPException` with corresponding error message.

**To Reproduce**
* `make install` 
* `make run-web-app`

- Filetype: image
- Any additional API parameters: `fast` strategy

Coniferish:
Closes #314

We don't support any fallback for fast strategy on images anymore, and unst will raise [value error](https://github.com/Unstructured-IO/unstructured/blob/main/unstructured/partition/strategies.py#L22) if partition uses fast strategy for image files. This catches the error and raises a 400 HTTPException.

Also, this refactors `pipeline_api` to extract out some validation of arguments.

### DIFF
--- a/prepline_general/api/general.py
+++ b/prepline_general/api/general.py
@@ -310,51 +310,20 @@ def pipeline_api(
 
         logger.debug(f"filetype: {file_content_type}")
 
-    # Reject traffic when free memory is below minimum
-    # Default to 2GB
-    mem = psutil.virtual_memory()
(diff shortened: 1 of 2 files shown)


awalker4 (approved):
Refactor looks great!

Issue #321, opened 2023-12-05
Support API Key for self-hosted 


omikader:
My team and I are using a self-hosted version of unstructured and we would like to secure our service, similar to how the hosted version uses API keys to prevent unauthorized access.

I was thinking it might be possible to pass in a single optional API key as an `ENV` variable to the Docker service such that any requests without the `unstructured-api-key` header are rejected. This way, we could reasonably ensure that our self-hosted version of unstructured is only invokable from our application.

Thanks for building such a great tool!

PR #322, opened 2023-12-05
feat: add support for API key validation for self-hosted

omikader:
This PR gives people the ability to use the API key header to validate requests when self-hosting Unstructured. If the optional environment variable of  `UNSTRUCTURED_API_KEY` does not match the request header of `unstructured-api-key`, then we fail with a `401` response instead of fulfilling the request.

This will allow people to self-host unstructured with confidence that only internal applications that have access to the shared key can use the service.

Closes https://github.com/Unstructured-IO/unstructured-api/issues/321

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.0.60
+
+* Enable self-hosted authorization using UNSTRUCTURED_API_KEY env variable
+
 ## 0.0.59
 
 * Bump unstructured to 0.11.0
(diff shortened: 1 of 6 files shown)


omikader:
@awalker4 all set! Thanks for the quick review. Let me know if there's anything else you'd like to see before merging 🙂

awalker4:
Final thing - can you bump the version to 0.0.60 in the changelog? You can run `make version-sync` to bump all the relevant files. Thanks!

PR #324, opened 2023-12-07
fix: handle argument of type 'int' is not iterable

### LINKED ISSUE #323: TypeError: argument of type 'int' is not iterable
Sentry Issue: [UNSTRUCTURED-API-AJ](https://unstructuredio.sentry.io/issues/4620178808/?referrer=github_integration)

```
TypeError: argument of type 'int' is not iterable

```

awalker4:
Closes #323 

Certain OSErrors come back with an errno in the first arg. We were searching for a substring in here, so we just need to skip the check if e.args[0] is not a string.

The new test case verifies this by raising an error with a errno.

### DIFF
--- a/prepline_general/api/general.py
+++ b/prepline_general/api/general.py
@@ -433,16 +433,17 @@ def pipeline_api(
             elements = partition(**partition_kwargs)
 
     except OSError as e:
-        if (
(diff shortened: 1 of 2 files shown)


yuming-long (approved):
LGTM!

PR #341, opened 2024-01-17
Klaijan/add error detail on GET request

Klaijan:
Add a new route handler for handling GET requests for the /general/v0/general. This handler will return the error message that informs user that GET request is not permitted.


awalker4 (approved):
LGTM!

PR #342, opened 2024-01-18
chore: revert base image change

awalker4:
The fix for openssl has been reverted: https://github.com/Unstructured-IO/base-images/pull/13

Need to confirm that this digest is still valid. If the docker test fails in this branch, we'll need to wait until the new image has been built. Edit: yeah, I've written over the existing base-image which may break self hosted users who are building from source with an older dockerfile. Will update the digest once the new image is out.


Coniferish (approved):
lgtm

PR #344, opened 2024-01-19
Set Speakeasy max retry time to 5 minutes

awalker4:
The clients are doing exponential backoff for up to an hour which is way too long. To verify, we'll need to wait for the client repos to get regenerated.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -22,7 +22,7 @@
         "backoff": {
             "initialInterval": 500,
             "maxInterval": 60000,
-            "maxElapsedTime": 3600000,
+            "maxElapsedTime": 300000,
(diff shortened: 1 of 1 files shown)


openapi.json
```diff
@@ -22,7 +22,7 @@
         "backoff": {
             "initialInterval": 500,
             "maxInterval": 60000,
-            "maxElapsedTime": 3600000,
+            "maxElapsedTime": 300000,
```
cragwolfe, line 25:
```suggestion
            "maxElapsedTime": 900000
```
actually how about 15 mins? :)

Coniferish (approved):
15 min back-off looks good

Issue #337, opened 2024-01-05
Chunking Parameters don't work as expected


JonasDupal:
**Describe the bug**
When setting the chunking_strategy to by_title and increasing the other chunking Parameters to values above the default, the results don't change. When setting the following values:
combine_under_n_chars = 1000000
new_after_n_chars = 1000000
max_characters = 1000000
multipage_sections = true
I would have expected to only get returned 1 section containing all the text of the files. But I get the same results as I did when not setting the parameters and using the default values, where I have multiple sections.
On the other hand when reducing the parameter values the results differ from the default.

**To Reproduce**
I am using Postman to test the API (https://api.unstructured.io/general/v0/general) and I am sending a pdf file

- Filetype: PDF
- Any additional API parameters: only setting the chunking_strategy parameters as described above

**Environment:**
 - Using the free API
 - Calling the API with postman

lambda-science:
Same issue.
With:
```python
params = {'skip_infer_table_types': '[]',
                         'chunking_strategy': 'by_title',
                         'combine_under_n_chars': '500',
                         'new_after_n_chars': '2000',
                         'max_characters': '2500',
                         'pdf_infer_table_structure': 'True',
                         'languages': 'eng',
                         'languages': 'fra',
                         'strategy': 'fast'}
```

I get elems that are like less than 100 caracters sometimes !

scanny:
@lambda-science the `combine_under_n_chars` argument can only reduce small chunks, it cannot eliminate them. So you can still see small chunks. The key constraint is that only consecutive chunks can be combined. Also, `Table` elements are never combined with any other element during chunking.

If you're seeing two consecutive chunks where the first is less than `combine_under_n_chars` and the second would fit together with the first in under `max_characters`, then that would indicate a problem.

Let us know if you're seeing that behavior and we can work to reproduce it. A PDF that reproduces the problem speeds things up considerably. Also a good first step is to see if you can reproduce it when partitioning/chunking locally, which would narrow the possible source of the problem.

lambda-science:
> @lambda-science the `combine_under_n_chars` argument can only reduce small chunks, it cannot eliminate them. So you can still see small chunks. The key constraint is that only consecutive chunks can be combined. Also, `Table` elements are never combined with any other element during chunking.
> 
> If you're seeing two consecutive chunks where the first is less than `combine_under_n_chars` and the second would fit together with the first in under `max_characters`, then that would indicate a problem.
> 
[...]

PR #345, opened 2024-01-19
fix(chunking): minimal fix for combine arg typo

scanny:
This is the minimal fix for the `combine_text_under_n_chars` arg-name typo bug described in #337.

Chunking parameters were added to the API about 4 months ago in this commit: https://github.com/Unstructured-IO/unstructured-api/commit/49fe71026e026530276bbb72f6a9df46feb6d1c2

At that time, the identifier `combine_under_n_chars` was used instead of `combine_text_under_n_chars` (as that parameter appears in `chunk_by_title()`). As a result, any value supplied for this argument via the REST API is ignored and defaults to 500. The `combine_under_n_chars` parameter name  appears in the API documentation and elsewhere (README.md etc.) and differs from the parameter name documented for `chunk_by_title()`.

A fuller remedy will be to add a separate `combine_text_under_n_chars` parameter to the API endpoint such that either name is accepted for now, then update the documentation such that all mentions are `combine_text_under_n_chars`. At some point in the future we may remove the `combine_under_n_chars` parameter from the endpoint.

However, making those changes requires a synchronized release of the `unstructured` library and in the interest of time we're making this "quick-fix" that can be released independently of any other dependencies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.0.63-dev0
+
+* Fix bug that ignored `combine_under_n_chars` chunking option argument.
+
 ## 0.0.62
 
 * Add hi_res_model_name to partition and deprecate model_name
--- a/prepline_general/api/app.py
+++ b/prepline_general/api/app.py
@@ -11,7 +11,7 @@ logger = logging.getLogger("unstructured_api")
 app = FastAPI(
     title="Unstructured Pipeline API",
     description="""""",
-    version="0.0.62",
+    version="0.0.63",
     docs_url="/general/docs",
     openapi_url="/general/openapi.json",
 )
--- a/prepline_general/api/general.py
+++ b/prepline_general/api/general.py
@@ -248,7 +248,7 @@ class ChipperMemoryProtection:
 
 def pipeline_api(
     file,
-    request=None,
+    request: Request,
     filename="",
     file_content_type=None,
     response_type="application/json",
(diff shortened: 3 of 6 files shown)


awalker4 (approved):
LGTM! Looks like that unit test I mentioned is catching this as expected. Parallel mode isn't so relevant right now, but that test is asserting that the api parameters are properly sent to partition, so I think you'll just need to rename the arg over there as well.

scanny:
> ... that test is asserting that the api parameters are properly sent to partition, so I think you'll just need to rename the arg over there as well.

The parameter name in `partition()` is correct, that's where the disconnect was. `combine_under_n_chars` just got lumped into `**kwargs` and hung around unused; `combine_text_under_n_chars` then took its default (500).

Actually, `partition()` doesn't define that parameter explicitly, it doesn't appear until `partition_pdf()` and even there it's not strictly required because `partition_pdf()` has `**kwargs` as well. It's the `@add_chunking_strategy()` decorator that actually goes looking for that specific name.

It's one of the challenges with taking advantage of the `**kwargs` behaviors, you lose type checking and then bugs like this one can find a home. I think the way I would solve this if we had the engineering will for it would be to create a `PartitionOptions` object that can be created at the earliest opportunity (by first receiver of call) and passed around whole from there on. I'd have to noodle it some more but for now I think we just have to be very diligent about arg names.

scanny:
@awalker4 how does the versioning work for this repo? Should I be bumping the version to `0.0.62-dev0` in `prepline_general/api/app.py` and add my `CHANGELOG` entry under that new heading?

If you can help me understand how this ends up appearing in production after merging, that would help me a lot so I can let the end-users know what to expect :)

awalker4:
@scanny The version change is driven by the changelog - so you can add a section there under 0.0.63-dev0. Running `make version-sync` will then replace the 0.0.62 string in the code. The endpoint will then live at `/general/v0/general` and `/general/v0.0.63/general`. Will dm you with deployment details.

PR #348, opened 2024-01-24
feat: hide GET helper handlers in swagger/schema

### LINKED ISSUE #347: API swagger docs should not display GET endpoints
Screenshot 2024-01-23 at 3 59 48 PM

We added GET handlers to return a useful error message. Need to tell FastAPI to remove them from the Swagger doc.

pawel-kmiecik:
Added `include_in_schema=False` in corresponding handlers

Screenshot 2024-01-24 at 16 00 01

Resolves #347


awalker4 (approved):
Thank you!!

PR #349, opened 2024-01-24
Add support for "basic" chunking and overlap

scanny:
**Summary**
- Draft that will probably be split into to separate PRs
- All but the last two commits are refactoring, groomed into coherent atomic commits focused on one refactoring aspect each.
- Core change is adding support for the new "basic" chunking strategy and the overlap features that come with it.


awalker4 (approved):
LGTM!

PR #350, opened 2024-01-24
enhancement: add yaml to default mime types

MthwRobinson:
### Summary

Related to https://github.com/Unstructured-IO/unstructured/pull/2446 . Adds YAML to the list of default MIME types. Had originally through to dynamically build the list from the MIME types in `unstructured`, but there are some MIME types (like `audio/wav` that are currently support for filetype detection but not `partition`. Probably want to be deliberate about adding those.


awalker4 (approved):
LGTM!

PR #353, opened 2024-01-25
Hide the versioned endpoint in the Swagger doc

awalker4:
The /general/v0.0.x/general path will change when we deploy a new version. This only serves to cause 404s for the users.


pawel-kmiecik (approved):
👍

PR #359, opened 2024-01-29
feat: roll openapi json into fastapi

pawel-kmiecik:
This PR:
- bumps `pydantic` to 2.5.x and removes it from explicit dependencies list (will be managed by fastapi)
- introduced Form params description in the code, which will form openapi and swagger documentation
- rolled back some openapi customizations
- keeps backward compatibility for passing parameters in form of `list[str]` (will not be shown in the documentation)


prepline_general/api/openapi.py
```diff
@@ -0,0 +1,111 @@
...
+    # and declare schema keys (type, format, description) as attributes
+    # https://fastapi.tiangolo.com/reference/openapi/models/?h=model
+    # Update the schema key from `Body_partition` to `partition_paramaters`
+
+    # Must manually correct the schema for the files parameter as due to a bug
+    # described here: https://github.com/tiangolo/fastapi/discussions/10280
+    # files parameter cannot be described with an annotation.
+    # TODO: Check if the bug is fixed and remove this workaround
+    for key in openapi_schema["components"]["schemas"]:
+        if "Body_general_partition_general_v0_general_post" in key:
```
pawel-kmiecik, line 83: That's something hard to avoid with app-defined parameters.

pawel-kmiecik:
@awalker4 should I remove `openapi.json` file from the repository then?

prepline_general/api/models/form_params.py
```diff
@@ -0,0 +1,209 @@
...
+    pdf_infer_table_structure: bool
+    strategy: str
+    extract_image_block_types: Optional[List[str]]
+    chunking_strategy: Optional[str]
+    combine_under_n_chars: Optional[int]
+    max_characters: int
+    multipage_sections: bool
+    new_after_n_chars: Optional[int]
+
+    @classmethod
```
awalker4, line 30: I take it this is for the backwards compatibility? Can we have a comment here to clarify how new params should be added?

prepline_general/api/openapi.py
awalker4: I suppose we can dig into later! It will already be a huge improvement to the swagger docs, so not too worried about the naming here yet.

awalker4:
I'm getting this error locally - any idea what happened?
Screenshot 2024-02-07 at 11 19 48 AM

awalker4:
> @awalker4 should I remove `openapi.json` file from the repository then?

Yes please!

prepline_general/api/models/form_params.py
pawel-kmiecik: for backward compatibility I added:
```
BeforeValidator(SmartValueParser[List[str]]().value_or_first_element),
```
In old fields that expected values both as a single value or as an element of a list.
The `SmartValueParser` will extract the first element from the list if the expected type is not a container.
`SmartValueParser` is a generic class and is parametrized with the expected type - in the above example its `List[str]`
You can check its tests to see how it's working:
[SmartValueParser tests](https://github.com/Unstructured-IO/unstructured-api/pull/359/files#diff-9c7f757ce9b6069a7eb77b85a1155564bb67e822d2fccd54e383c1714429c755)
[Deprecated API tests (with params passed as lists)](https://github.com/Unstructured-IO/unstructured-api/pull/359/files#diff-465e31562cbb79b8c3177e91026b46eb3f2edc6711d48f462e1f470f528856c0)

awalker4 (approved):

pawel-kmiecik:
> I'm getting this error locally - any idea what happened? Screenshot 2024-02-07 at 11 19 48 AM

Should be fixed now, I removed `openapi.json` static file as well.

PR #363, opened 2024-02-06
test: test extract_image_block_type parameter

ds-filipknefel:
Test if image metadata image_base64 and image_mime_type are present when extract_image_block_type parameter is given.


christinestraub (approved):
LGTM!

PR #367, opened 2024-02-14
Update readme note on minimum free memory check

awalker4:
Update the wording to:
* Clarify that this is an environment variable
* Explain how server behavior changes when the value is changed


yuming-long (approved):
LGTM

PR #368, opened 2024-02-15
feat: support list type parameters

christinestraub:
The purpose of this PR is to support parsing all list type parameters, including `extract_image_block_types` when calling unstructured API via unstructured client SDK (Python/JS) generated by `speakeasy`. 

Currently, the `speakeasy` doesn't generate proper client code to pass list type parameters to unstructured API because they do not expect to support specific client code for `FastAPI` that the unstructured API relies on. To address this issue, I updated the unstructured API code to parse all list type parameters passed as JSON-formatted lists (e.g. `'["image", "table"]'`).

**NOTE:** You must pass the list type parameter as a JSON-formatted list when calling unstructured API via unstructured client SDK.
(e.g. `extract_image_block_types = '["image", "table"]'`, `skip_infer_table_types='["docx", "xlsx"]'`...) 

### Summary
- update `SmartValueParser.value_or_first_element()` to parse JSON format string (e.g. `'["image", "table"]'`) that is convertible to a list
- apply `SmartValueParser.value_or_first_element()` to all list type parameters
- remove existing `extract_image_block_types` parsing logic

### Testing
- via unstructured_client_sdk (Python)
```
s = UnstructuredClient(
    server_url="http://localhost:8000/general/v0/general",
    api_key_auth="YOUR-API-KEY"
)

filename = "sample-docs/embedded-images-tables.pdf"

with open(filename, "rb") as f:
    # Note that this currently only supports a single file
    files = shared.Files(
        content=f.read(),
        file_name=filename,
    )

req = shared.PartitionParameters(
    files=files,
    # Other partition params
    strategy="hi_res",
    extract_image_block_types='["image", "table"]',
    languages=["pdf"],
)

try:
    resp = s.general.partition(req)
    print([el.get("metadata").get("image_mime_type") for el in resp.elements if el.get("metadata").get("image_mime_type")])
except SDKError as e:
    print(e)
```

- via unstructured_client_sdk (JS)
```
import { UnstructuredClient } from "unstructured-client";
import * as fs from "fs";

const key = "YOUR-API-KEY";

const client = new UnstructuredClient({
    serverURL: "http://localhost:8000",
    security: {
        apiKeyAuth: key,
    },
});

const filename = "sample-docs/embedded-images-tables.pdf";
const data = fs.readFileSync(filename);

client.general.partition({
    // Note that this currently only supports a single file
    files: {
        content: data,
        fileName: filename,
    },
    // Other partition params
    strategy: "hi_res",
[...]

PR #370, opened 2024-02-21
 chore: change table extraction defaults

ds-filipknefel:
Change default values for table extraction - relies on [this](https://github.com/Unstructured-IO/unstructured/compare/chore/change-table-extraction-defaults?expand=1) `unstructured` PR being merged and synced.

We want to move away from `pdf_infer_table_structure` parameter, in this PR we:

- [x] set `pdf_infer_table_structure=True` and `skip_infer_table_types=[]` by default
- [x] remove it from the examples in documentation
- [x]  describe it as deprecated in favor of `skip_infer_table_types` in documentation
- [x] update the tests to reflect new usage
- [x] bump `unstructured` dependecy to at least 0.13.0 (REQUIRED)

More detailed description of how we want parameters to interact
- if `pdf_infer_table_structure` is False tables will never extracted from pdf
- if `pdf_infer_table_structure` is True tables will be extracted from pdf unless it's skipped via `skip_infer_table_types`
- on default `pdf_infer_table_structure=True` and `skip_infer_table_types=[]`


yuming-long:
I thought adding table extraction would increase the API response time. Mind if i ask why we are setting this default to true?

ds-filipknefel:
> I thought adding table extraction would increase the API response time. Mind if i ask why we are setting this default to true?

On the request of @pawel-kmiecik, can you chime in for the reasons for it?

pawel-kmiecik:
> > I thought adding table extraction would increase the API response time. Mind if i ask why we are setting this default to true?
> 
> On the request of @pawel-kmiecik, can you chime in for the reasons for it?

That's what I set with @cragwolfe in private conversation about parameters steering the table extraction:
> e.g. with curl, one should just now just use the parameter:
> -F skip_infer_table_types="[]"
> to always get tables extracted if they are available
> for the API

AFAIK both `pdf_infer_table_structure="true"` and `skip_infer_table_types="[]"` must be explicitly set by a user to get the table extraction. We'd like to leave only `skip_infer_table_types="[]"`

yuming-long:
gotcha, thanks!

ds-filipknefel:
> > > I thought adding table extraction would increase the API response time. Mind if i ask why we are setting this default to true?
> > 
> > 
> > On the request of @pawel-kmiecik, can you chime in for the reasons for it?
> 
> That's what I set with @cragwolfe in private conversation about parameters steering the table extraction:
> 
> > e.g. with curl, one should just now just use the parameter:
> > -F skip_infer_table_types="[]"
> > to always get tables extracted if they are available
> > for the API
> 
> AFAIK both `pdf_infer_table_structure="true"` and `skip_infer_table_types="[]"` must be explicitly set by a user to get the table extraction. We'd like to leave only `skip_infer_table_types="[]"`

Actually right now it's enough to set `pdf_infer_table_structure=true` because of how it's [handled](https://github.com/Unstructured-IO/unstructured/blob/4d12c61cb83c559cc795b260177d9562532077a9/unstructured/partition/auto.py#L556) internally. Logging will throw a warning and tell you to correct yourself but actually ignore `skip_infer_table_types` contents. 

So with this change if user leaves `pdf_infer_table_structure` or `skip_infer_table_types` with default values, extraction will happen.

README.md
```diff
@@ -131,21 +131,6 @@ When elements are extracted from PDFs or images, it may be useful to get their b
...
-  -H 'Content-Type: multipart/form-data' \
-  -F 'files=@sample-docs/layout-parser-paper.pdf' \
-  -F 'strategy=hi_res' \
-  -F 'pdf_infer_table_structure=true' \
-  | jq -C . | less -R
-```
-
 #### Skip Table Extraction
 
 Currently, we provide support for enabling and disabling table extraction for file types other than PDF files. Set parameter `skip_infer_table_types` to specify the document types that you want to skip table extraction with. By default, we skip table extraction
```
yuming-long, line 136: here `By default, we skip table extraction`.... i realize that the default here will be change to perform table extraction for all file types (i assume this is the case?) Could you update it as well? thx

yuming-long: ohhh right sorry i was confused

yuming-long: so default is to skip every file expect pdf for hi res table extraction?

cragwolfe: that used to be the default, but let's change it to not skip any file types (empty list value for `skip_infer_table_types`). the cost for table extraction is not too high.

moving forward, the preferred way to skip table extraction for pdfs, png, and jpg (which adds a little of compute cost) would be:
skip_infer_table_types=["png","jpg","pdf"]

^^ the values to skip_infer_table_types should be correct json per https://github.com/Unstructured-IO/unstructured-api/pull/368

pawel-kmiecik: @ds-filipknefel ok as this is a really small change - please remove the contents of the default list here:
https://github.com/Unstructured-IO/unstructured-api/blob/main/prepline_general/api/models/form_params.py#L75
and from the parameter description. 

Only leave `"['pdf', 'jpg', 'png']"` in the example param.
If any readmes need to be updated - please do it as well.

yuming-long (approved):
LGTM!

PR #372, opened 2024-02-22
test: extract_image_block_types parameter usage

christinestraub:
This PR updates the unit test function to test if image metadata `image_base64` and `image_mime_type` are present when `extract_image_block_types` parameter is given.


ds-filipknefel:
LGTM but I don't know if it's necessary or if that's the right place for this

How these arguments are parsed is entirely up to `Request` class so maybe we just write a single test which shows that passing either one of `["str1", "str2"]` or `'["str1", "str2"]'` as a value is equivalent and do not tie it to specific usages?

Following this PR we should probably update `def test_ocr_languages_param()` as well for consistancy because it also takes a parameter with a list as value.

I'd expect different parameters in this test to differ in contents of the array not how that array is represented.

christinestraub:
I think it's necessary and we need to make sure that unstructured API can parse different input formats for same list parameter (e.g. ["str1", "str2"] or '["str1", "str2"]')

For `test_ocr_languages_param()`,  I updated the function in PR #375.

PR #373, opened 2024-02-22
fix: receive csv when `output_format=text/csv`

ds-filipknefel:
Issues with not receiving csv formatted answer stem from looking up `accept:` headers to derive in what format we want to send a response.

I suggest that `output_format` should be wholly in charge of this and that we get rid of raising `406` errors when there's a mismatch between accept headers and response type as user explicitly knows what format they'll get.

The only exception for it is for `multipart/mixed` but this solution does everything the same when `accept: multipart/mixed` is provided (which right now I'm not sure works over all, I didn't manage to successfully send a request through curl with this header included)

### DIFF
--- a/prepline_general/api/general.py
+++ b/prepline_general/api/general.py
@@ -751,12 +751,6 @@ def general_partition(
                 files[file_index], form_params.gz_uncompressed_content_type
             )
 
-    default_response_type = form_params.output_format or "application/json"
-    if not content_type or content_type == "*/*" or content_type == "multipart/mixed":
-        media_type = default_response_type
-    else:
-        media_type = content_type
-
     def response_generator(is_multipart: bool):
         for file in files:
             file_content_type = get_validated_mimetype(file)
@@ -775,7 +769,7 @@ def general_partition(
                 skip_infer_table_types=form_params.skip_infer_table_types,
                 strategy=form_params.strategy,
                 xml_keep_tags=form_params.xml_keep_tags,
-                response_type=media_type,
+                response_type=form_params.output_format,
                 filename=str(file.filename),
                 file_content_type=file_content_type,
                 languages=form_params.languages,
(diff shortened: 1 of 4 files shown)


prepline_general/api/general.py
```diff
@@ -775,7 +769,7 @@ def response_generator(is_multipart: bool):
                 skip_infer_table_types=form_params.skip_infer_table_types,
                 strategy=form_params.strategy,
                 xml_keep_tags=form_params.xml_keep_tags,
-                response_type=media_type,
+                response_type=form_params.output_format,
```
pawel-kmiecik, line 772: The code is simplified now but I think we should limit the possibilities for `form_params.output_format` to a list of acceptable output formats, like it's done with `Literal` here:
https://github.com/Unstructured-IO/unstructured-api/blob/main/prepline_general/api/models/form_params.py#L138

This would require updating the doc as well. Not sure if we need to do anything in `openapi.json` static file then and also consult with `python_client` repo maintainers (cc: @hubert-rutkowski85 )

Except that one - looks ok 👍

hubert-rutkowski85: Thanks for mentioning. I agree about adding Literal to make it more specific and less error prone.

As for python-client, it's defined there as:
```
output_format: Optional[str] = dataclasses.field(default=None, metadata={'multipart_form': { 'field_name': 'output_format' }})
r"""The format of the response. Supported formats are application/json and text/csv. Default: application/json."""
```
It's code generated from openapi.json, where the `output_format` is defined as str. This means that fortunately we don't need to change anything in client now. It's just backend spec that becomes more defined in what it accepts. In the future enums could be used to make openapi.json more specified also on client side, but that's better when done all at once and in separate ticket.

ds-filipknefel: Changed the type to `Literal`

pawel-kmiecik (approved):
LGTM 👍

yuming-long (commented):
Please update tests with csv conversion (`pd.read_csv`) for example [here](https://github.com/Unstructured-IO/unstructured-api/blob/main/test_general/api/test_app.py#L102) (another one is in smoke test). I don't think we need this if csv is supported. Otherwise LGTM!

ds-filipknefel:
> Please update tests with csv conversion (`pd.read_csv`) for example [here](https://github.com/Unstructured-IO/unstructured-api/blob/main/test_general/api/test_app.py#L102) (another one is in smoke test). I don't think we need this if csv is supported. Otherwise LGTM!

I actually think using `pd.read_csv` is good here and would rewrite my tests to use it as well, given that our csv does describe a table.
If we operate on csv str to e.g. check number of elements we can run into trouble, I've actually run into that right now when trying to remove `pd.read_csv`. A single value in csv represented table can span multiple lines so we can't rely on counting lines to validate number of elements.

Letting pandas parse this csv into a table makes it easier.

PR #396, opened 2024-03-21
feat: add include_orig_elements chunking parameter

scanny:
**Summary**
Add the new `include_orig_elements` chunking parameter so it appears in the SDKs.


awalker4 (approved):
LGTM!

PR #399, opened 2024-03-26
feat: add unique_element_ids param

hubert-rutkowski85:
Api gets the `unique_element_ids` added as a parameter, and passed into the library.

You can test with `make test`.


cragwolfe (commented):
Can we indicate in the docs that that "unique" here means "random"?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Add support for both `list[str]` and `str` input formats for `ocr_languages` parameter
 * Adds support for additional MIME types from `unstructured`
 * Document the support for gzip files and add additional testing 
+* Add support for `unique_element_ids` parameter.
```
awalker4, line 7: Can you start a 0.0.66-dev and run make version-sync here

hubert-rutkowski85: Sure!

hubert-rutkowski85:
@cragwolfe 
> Can we indicate in the docs that that "unique" here means "random"?

After "_Different elements with the same text will have the same ID, and there could also be hash collisions.
To use UUIDs in the output instead, set unique_element_ids=true_"

Added "_Note: this means that the element IDs will be random, so with every partition of the same file, you will get different IDs._ "

Also bumped version as @awalker4 asked. Basically **the PR is ready to merge**, the CI troubles cloud that all tests pass (locally) and they did pass with previous commits, and there were no quality code changes since then. So if you are ok with it, I'd want to merge to avoid having long living PRs, which need to be continously updated (merges from main) and risk conflicts.

hubert-rutkowski85:
Ok, tests are passing now, @cragwolfe @awalker4 please approve.

PR #400, opened 2024-04-04
Support for `starting_page_number` parameter when doing PDF page splitting

micmarty-deepsense:
This PR enables the Python and JS clients to partition PDF pages independently after splitting them on their side (`split_pdf_page=True`). Splitting is also supported by API itself - this makes sense when users send their requests without using our dedicated clients.

Related to: 
* https://github.com/Unstructured-IO/unstructured/pull/2842
* https://github.com/Unstructured-IO/unstructured/pull/2673

It should be merged before these:
* https://github.com/Unstructured-IO/unstructured-js-client/pull/55
* https://github.com/Unstructured-IO/unstructured-python-client/pull/72

**The tests for this PR won't pass until the related PRs are both merged.**

## How to test it locally
Unfortunately the `pytest` test is not fully implemented, it fails - see [this comment](https://github.com/Unstructured-IO/unstructured-api/pull/400#discussion_r1581306781)
1. Clone Python client and checkout to this PR: https://github.com/Unstructured-IO/unstructured-js-client/pull/55
2. `cd unstructured-client; pip install --editable .`
3. `make run-web-app`
4.  `python <script-below>.py`

```python
from unstructured_client import UnstructuredClient
from unstructured_client.models import shared
from unstructured_client.models.errors import SDKError

s = UnstructuredClient(api_key_auth=os.environ["UNS_API_KEY"], server_url="http://localhost:8000")

# -- this file is included in this PR --
filename = "sample-docs/DA-1p-with-duplicate-pages.pdf"
with open(filename, "rb") as f:
    files = shared.Files(content=f.read(), file_name=filename)

req = shared.PartitionParameters(
    files=files,
    strategy="fast",
    languages=["eng"],
    split_pdf_page=False, # this forces splitting on API side (if parallelization is enabled)
    # split_pdf_page=True,  # forces client-side splitting, implemented here: https://github.com/Unstructured-IO/unstructured-js-client/pull/55
)
resp = s.general.partition(req)
ids = [e["element_id"] for e in resp.elements]
page_numbers = [e["metadata"]["page_number"] for e in resp.elements]
# this PDF contains 3 identical pages, 13 elements each
assert page_numbers == [1,1,1,1,1,1,1,1,1,1,1,1,1,
                        2,2,2,2,2,2,2,2,2,2,2,2,2,
                        3,3,3,3,3,3,3,3,3,3,3,3,3]
assert len(ids) == len(set(ids)), "Element IDs are not unique"
```


prepline_general/api/general.py
```diff
@@ -225,12 +227,16 @@ def partition_pdf_splits(
     results: List[Element] = []
     page_iterator = get_pdf_splits(pdf_pages, split_size=pages_per_pdf)
 
+    # -- retain the original value for the next step of assembling page splits. --
+    unique_element_ids = partition_kwargs.pop("unique_element_ids")
```
cragwolfe, line 231: we don't need this anymore, do we, since we'll get this from unstructured now?

micmarty-deepsense: that's right, I'm reworking this PR to utilize `starting_page_number` instead

test_general/api/test_app.py
```diff
@@ -553,6 +575,72 @@ def json(self):
...
+    client = TestClient(app)
+    monkeypatch.setattr(
+        general,
+        "call_api",
+        lambda *args, **kwargs: call_api_using_test_client(*args, client=client, **kwargs),
+    )
+
+    # -- there are 3 pages identical pages in this pdf --
+    test_file = Path("sample-docs") / "DA-1p-with-duplicate-pages.pdf"
+    response = client.post(
```
micmarty-deepsense, line 618: during testing we want to use TestClient from FastAPI, but the way splitting is implemented relies on sending API request with PDF fragments to itself (recursively, kind of). So to inject the test client I tried to mock it using `call_api_using_test_client`, expecting that everything will work fine. Unfortunately this test fails, and page_numbers are not assigned correctly...

awalker4 (approved):

PR #402, opened 2024-04-10
chore(deps): bump unstructured to 0.13.2

ds-filipknefel:
Bump `unstructured` dependency to 0.13.2

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -63,7 +63,6 @@ filelock==3.13.4
     #   huggingface-hub
     #   torch
     #   transformers
-    #   triton
 filetype==1.2.0
     # via unstructured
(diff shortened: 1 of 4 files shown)


requirements/base.txt
```diff
@@ -150,11 +148,42 @@ numpy==1.26.4
     #   torchvision
     #   transformers
     #   unstructured
+nvidia-cublas-cu12==12.1.3.1
```
awalker4, line 151: These packages if I remember right show up when we pip compile on Linux rather than Mac. Can you rerun on a mac?

yuming-long (approved):
LGTM

yuming-long (approved):
LGTM

PR #405, opened 2024-05-01
fix/Update parallel mode test to fix CI

awalker4:
Some of the output has changed in this test and it's breaking the docker build action.

https://github.com/Unstructured-IO/unstructured-api/actions/runs/8898982407/job/24437478433


yuming-long (approved):
LGTM

PR #414, opened 2024-05-11
fix/skip another test pdf on linux/arm64

awalker4:
Follow up from #413. These tests hang in CI and are blocking the image build.


ryannikolaidis (approved):

awalker4:
Thank you @ryannikolaidis!

PR #417, opened 2024-05-23
API refactor proposition

hubert-rutkowski85:
It aims to make the code structure easier to understand, navigate and modify, by:

1. shortening long files - `general.py` had 849 lines, now 553. 

2. moving common functionality to seperate files - endpoints.py, logging.py, validation.py, memory_protection.py  

In both cases, it still could be improved (general.py shorteneded by about 300 lines, and functions moved to new files like pdf_splits) , but had some problems with monkeypatching expecting functions in certain places, and didn't want to spend more time on it.

### DIFF
--- a/prepline_general/api/endpoints.py
+++ b/prepline_general/api/endpoints.py
@@ -163,3 +163,8 @@ async def handle_invalid_get_request():
     raise HTTPException(
         status_code=status.HTTP_405_METHOD_NOT_ALLOWED, detail="Only POST requests are supported."
     )
+
+
+@router.get("/healthcheck", status_code=status.HTTP_200_OK, include_in_schema=False)
+def healthcheck(request: Request):
+    return {"healthcheck": "HEALTHCHECK STATUS: EVERYTHING OK!"}
--- a/prepline_general/api/endpoints.py
+++ b/prepline_general/api/endpoints.py
@@ -6,7 +6,7 @@ import os
 from typing import List, Sequence, Dict, Any, cast, Union
 
 import pandas as pd
-from fastapi import FastAPI, APIRouter, UploadFile, Depends, HTTPException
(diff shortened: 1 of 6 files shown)


cragwolfe (commented):
i'd feel better about a large refactor change like thais after we had a multi-thousand doc release candidate test in place, since basically right now, prod is inadvertently serving as our QA env as well.
not saying don't do it, but need to be super cautious.

prepline_general/api/endpoints.py
```diff
@@ -0,0 +1,169 @@
...
+        )
+        if len(responses) > 1:
+            for resp in responses[1:]:
+                resp_data = pd.read_csv(  # pyright: ignore[reportUnknownMemberType]
+                    io.BytesIO(resp.body)
+                )
+                data = data.merge(  # pyright: ignore[reportUnknownMemberType]
+                    resp_data, how="outer"
+                )
+        return PlainTextResponse(data.to_csv())
```
mpolomdeepsense, line 144: These two function could potentially be moved out of this function. Would require passing some more arguments but I think it would look cleaner.

hubert-rutkowski85:
> i'd feel better about a large refactor change like thais after we had a multi-thousand doc release candidate test in place, since basically right now, prod is inadvertently serving as our QA env as well. not saying don't do it, but need to be super cautious.

Yes I agree, having tests when making refactors is crucial. Here the situation is simpler, as the only changes are moving functions between files. So no changes in logic. All existing tests pass.

prepline_general/api/endpoints.py
hubert-rutkowski85: Right, good point.

hubert-rutkowski85: Done.

awalker4 (approved):

PR #419, opened 2024-05-31
chore: Remove the create_issue workflow

awalker4:
We no longer need to mirror gh issues in the internal jira.


MthwRobinson (approved):
LGTM!

PR #420, opened 2024-06-05
build(deps): version bumps for maintenance

MthwRobinson:
### Summary

Version bumps for regular maintenance and to address moderate CVEs from security scans. Also updates the `unstructured` extra from `local-inference` to `all-docs` to keep up with latest best practices for the `unstructured` library.

Includes an update for appropriately setting `pdf_infer_table_structure` depending on the value of `skip_infer_table_types` and adds a test.


cragwolfe (approved):

prepline_general/api/general.py
```diff
@@ -295,7 +295,7 @@ def pipeline_api(
     hi_res_model_name: Optional[str] = None,
     include_page_breaks: bool = False,
     ocr_languages: Optional[List[str]] = None,
-    pdf_infer_table_structure: bool = True,
+    pdf_infer_table_structure: bool = False,
```
MthwRobinson, line 298: @awalker4 - Before I merge here, could you check my math on this? Had to make this change for `scripts/smoketest.py` to pass because the default in `unstructured` changed to "table extraction off for PDFs and images". With this update, the default API behavior doesn't change for `skip_infer_table_types`, which is now the supported parameter for changing table behavior. It does change for `pdf_infer_table_structure`, which we've been emitting deprecation warnings about for a while. Does that make sense?

MthwRobinson: Disregard this comment

awalker4 (approved):

awalker4:
Oh wait can you also bump the version and note the unstructured version in the changelog?

PR #425, opened 2024-06-20
fix: build and push workflow failing due to missing `-f` option `buildx build` command

micmarty-deepsense:
I noticed that images on main branch are failing to build (and push) due to missing `-f` parameter in `docker buildx build`. By default it expects `Dockerfile` to exist, but we only have `Dockerfile-amd64` and `Dockerfile-arm64`

image

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -69,7 +69,7 @@ jobs:
         sudo rm -rf /usr/share/dotnet /opt/ghc /usr/local/share/boost
 
         ARCH=$(cut -d "/" -f2 <<< ${{ matrix.docker-platform }})
-        DOCKER_BUILDKIT=1 docker buildx build --platform=$ARCH --load \
+        DOCKER_BUILDKIT=1 docker buildx build --platform=$ARCH --load -f Dockerfile-$ARCH \
           --build-arg PIP_VERSION=$PIP_VERSION \
           --build-arg BUILDKIT_INLINE_CACHE=1 \
           --build-arg PIPELINE_PACKAGE=${{ env.PIPELINE_FAMILY }} \
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -76,8 +76,8 @@ jobs:
           --build-arg PIPELINE_PACKAGE=${{ env.PIPELINE_FAMILY }} \
           --provenance=false \
           --progress plain \
-          --cache-from $DOCKER_BUILD_REPOSITORY:$ARCH \
-          -t $DOCKER_BUILD_REPOSITORY:$ARCH-$SHORT_SHA .
+          --cache-from $DOCKER_BUILD_REPOSITORY:${{ matrix.arch }} \
+          -t $DOCKER_BUILD_REPOSITORY:${{ matrix.arch }}-$SHORT_SHA .
     - name: Set virtualenv cache
       uses: actions/cache@v4
       id: virtualenv-cache
@@ -99,7 +99,7 @@ jobs:
     - name: Push image
       run: |
         # write to the build repository to cache for the publish-images job
-        docker push $DOCKER_BUILD_REPOSITORY:$ARCH-$SHORT_SHA
+        docker push $DOCKER_BUILD_REPOSITORY:${{ matrix.arch }}-$SHORT_SHA
   publish-images:
     runs-on: ubuntu-latest-m
     needs: [setup, set-short-sha, build-images]
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -45,16 +45,17 @@ jobs:
   build-images:
     strategy:
       matrix:
-        docker-platform: ["linux/arm64", "linux/amd64"]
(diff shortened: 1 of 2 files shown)


MthwRobinson (approved):
LGTM

.github/workflows/docker-publish.yml
```diff
@@ -68,8 +69,8 @@ jobs:
         # Clear some space (https://github.com/actions/runner-images/issues/2840)
         sudo rm -rf /usr/share/dotnet /opt/ghc /usr/local/share/boost
 
-        ARCH=$(cut -d "/" -f2 <<< ${{ matrix.docker-platform }})
-        DOCKER_BUILDKIT=1 docker buildx build --platform=$ARCH --load \
+        DOCKER_BUILDKIT=1 docker buildx build --load -f Dockerfile-${{ matrix.arch }} \
```
micmarty-deepsense, line 72: this single line fixes the issue, the rest is refactor

.github/workflows/docker-publish.yml
```diff
@@ -68,8 +69,8 @@ jobs:
...
+          --platform=$DOCKER_PLATFORM \
```
micmarty-deepsense, line 73: technically this should be e.g. "linux/amd64", but "amd64" (as it was) would also work

.github/workflows/docker-publish.yml
```diff
@@ -88,19 +89,16 @@ jobs:
       uses: docker/setup-qemu-action@v3
     - name: Test image
       run: |
-        ARCH=$(cut -d "/" -f2 <<< ${{ matrix.docker-platform }})
         source .venv/bin/activate
-        if [ "${{ matrix.docker-platform }}" == "linux/arm64" ]; then
-          DOCKER_PLATFORM="${{ matrix.docker-platform }}" DOCKER_IMAGE="$DOCKER_BUILD_REPOSITORY:$ARCH-$SHORT_SHA" \
+        DOCKER_IMAGE="$DOCKER_BUILD_REPOSITORY:${{ matrix.arch }}-$SHORT_SHA" \
+        if [ "$DOCKER_PLATFORM" == "linux/arm64" ]; then
```
micmarty-deepsense, line 94: DOCKER_PLATFORM env variable is already set, so no need to provide it here. If you expect being explicit, please let me know

Dockerfile-amd64
```diff
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:experimental
-FROM quay.io/unstructured-io/base-images:wolfi-base@sha256:6c00a236c648ffdaf196ccbc446f5c6cc9eb4e3ab9e437178abcfac710b2b373 as base
```
micmarty-deepsense, line 2: this image does not exist anymore, I had to update SHA

micmarty-deepsense:
FYI This PR is blocked by some random issue with apt repository being not responsive (timeout)
https://github.com/Unstructured-IO/unstructured-api/actions/runs/9598390479/job/26473331895?pr=425

christinestraub (changes requested):
Is the `ARCH` env variable also already set or do we need to replace it with {{ matrix.arch }}?

micmarty-deepsense:
> Is the `ARCH` env variable also already set or do we need to replace it with {{ matrix.arch }}?

It's no longer being set, but technically we could, it depends what we prefer. We could go with this:
```yml


...
somejob:
    env:
        ARCH: ${{ matrix.arch }}
    run: echo $ARCH
``` 
but IMO it's better to stick with only one way of accessing the value we want to use.

My primary goal was to eliminate redundant usages of `cut` and `fd` in many places. Even though we all understand what the command does, it's unnecessarily "complex". It's easier to have "amd64" matrix value and prepend it with "linux/" in env var. WDYT? @christinestraub

christinestraub:
> It's no longer being set, but technically we could, it depends what we prefer. We could go with this:
> 
> ```yaml
> ...
> somejob:
>     env:
>         ARCH: ${{ matrix.arch }}
>     run: echo $ARCH
> ```
> 
> but IMO it's better to stick with only one way of accessing the value we want to use.
> 
> My primary goal was to eliminate redundant usages of `cut` and `fd` in many places. Even though we all understand what the command does, it's unnecessarily "complex". It's easier to have "amd64" matrix value and prepend it with "linux/" in env var. WDYT? @christinestraub

Sounds reasonable!

micmarty-deepsense:
> > It's no longer being set, but technically we could, it depends what we prefer. We could go with this:
> > ```yaml
> > ...
> > somejob:
> >     env:
> >         ARCH: ${{ matrix.arch }}
> >     run: echo $ARCH
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but IMO it's better to stick with only one way of accessing the value we want to use.
> > My primary goal was to eliminate redundant usages of `cut` and `fd` in many places. Even though we all understand what the command does, it's unnecessarily "complex". It's easier to have "amd64" matrix value and prepend it with "linux/" in env var. WDYT? @christinestraub
> 
> Sounds reasonable!

if everything else looks good, please approve

christinestraub (approved):

PR #426, opened 2024-06-20
build: update amd64 wolfi base image

christinestraub:


christinestraub:
Addressed by https://github.com/Unstructured-IO/unstructured-api/pull/427

PR #455, opened 2024-08-22
feat/add include_slide_notes parameter

mackurzawa:
## Description
* Added `include_slide_notes` parameter, default is `True`. Works for `.ppt` and `.pptx` file extensions.
* Added two new files in `sample-docs`: `sample-docs/notes.ppt`, `sample-docs/notes.pptx` that include notes on their slides. This is to easily test the functionality, as there are no existing PowerPoint files that include slide notes.

## Testing
```
#  using default value (True) returns additional NarrativeText element that contains notes
curl -X 'POST'   'http://localhost:8000/general/v0/general'   -H 'accept: application/json'   -H 'Content-Type: multipart/form-data'   -F 'files=@sample-docs/notes.pptx'   -F 'output_format="text/csv"' 

# explicit include_slide_notes=True returns additional NarrativeText element that contains notes
curl -X 'POST'   'http://localhost:8000/general/v0/general'   -H 'accept: application/json'   -H 'Content-Type: multipart/form-data'   -F 'files=@sample-docs/notes.pptx'   -F 'output_format="text/csv"' -F 'include_slide_notes=True'

# explicit include_slide_notes=False returns no NarrativeText element 
curl -X 'POST'   'http://localhost:8000/general/v0/general'   -H 'accept: application/json'   -H 'Content-Type: multipart/form-data'   -F 'files=@sample-docs/notes.pptx'   -F 'output_format="text/csv"' -F 'include_slide_notes=False'
```

Same with file `notes.ppt`


mackurzawa:
Hey @awalker4,
Do you think it is worth mentioning this change in the changelog as a breaking change due to the different default behavior? Here’s why I’m wondering: if a user calls the API with a PowerPoint file that includes slide notes, they will receive additional elements related to those notes. While the previous elements will remain unchanged, their order might be different. I think it’s pretty rare for a user to care about the element order, but still, it might be worth noting

awalker4 (approved):
LGTM! On the element order differences, I think it's worth adding another sentence in the changelog, but that's probably all I'd do in this case. Thanks!

PR #457, opened 2024-09-09
fix: improve input validation to allow quotes for strategy parameter

jordan-homan:


awalker4 (approved):
LGTM! Can you add a note at the top of changelog.md? I believe the last release was just cut, in which case you can:

- Start a new section with 0.0.xx-dev0
- Run `make version-sync` to copy the new version string into the code. You can use `make check` to verify that it's happy.

awalker4 (approved):

PR #486, opened 2025-01-23
Update app-start.sh

Umaaz:
add option to set HOST for uvicorn start command

We want to control the host for our deployment.


awalker4 (approved):
Thanks!

PR #499, opened 2025-04-22
Add support for 'basic' chunking strategy to match documentation

Sameena-Thabassum:
**Problem Statement**
The unstructured-api currently lacks support for the 'basic' chunking strategy, which is mentioned in the official documentation. As per the documentation, the API should allow users to choose between different chunking strategies, such as "by_title" and "basic". However, the chunking_strategy parameter in the API currently only supports 'by_title', causing a mismatch between the functionality described in the documentation and the actual behavior of the system.

**Solution**
Updating the chunking_strategy parameter: The chunking_strategy parameter is modified to support both 'by_title' and 'basic' chunking strategies, as indicated by the official documentation.

The Literal type hint for chunking_strategy is updated to include both 'by_title' and 'basic' as valid options.

This update ensures that the API accepts 'basic' as a valid chunking strategy and prevents potential issues with incorrect values.

**Why this PR should be accepted**
This pull request resolves a key issue by ensuring that the unstructured-api behaves consistently with the documented features, particularly by adding support for the 'basic' chunking strategy.

### DIFF
--- a/prepline_general/api/models/form_params.py
+++ b/prepline_general/api/models/form_params.py
@@ -178,11 +178,11 @@ class GeneralFormParams(BaseModel):
         ] = False,
         # -- chunking options --
         chunking_strategy: Annotated[
-            Optional[Literal["by_title"]],
+            Optional[Literal["by_title", "basic"]],
             Form(
                 title="Chunking Strategy",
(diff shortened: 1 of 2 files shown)


Sameena-Thabassum:
@dkarlovi Please review

PR #500, opened 2025-04-25
pip install due to cve for h11

potter-potter:
cve scan found vulnerability in h11. supposedly 0.16.0 fixes this.


awalker4:
Looks like we can close in favor of https://github.com/Unstructured-IO/unstructured-api/pull/502

PR #509, opened 2025-06-05
parallelize tests

jiajun-unstructured:
This repo doesn't have access to the self-hosted-xlarge runner for more cores, so this change will only start 2 workers


requirements/test.txt
```diff
@@ -421,15 +430,15 @@ nltk==3.9.1
...
+notebook==7.4.3
     # via jupyter
 notebook-shim==0.2.4
     # via
     #   jupyterlab
     #   notebook
 numpy==1.26.4
     # via
-    #   -c requirements/constraints.in
+    #   -c /Users/jiajunxu/code/unstructured-api/requirements/constraints.in
```
PastelStorm, line 441: pip-compile loves to add absolute paths in the comments, so you need to delete them all :D

PastelStorm (changes requested):

PastelStorm (approved):

PastelStorm:
> This repo doesn't have access to the self-hosted-xlarge runner for more cores, so this change will only start 2 workers

for public repos the number of cores is actually 4, which I think is good enough.

PastelStorm:
> This repo doesn't have access to the self-hosted-xlarge runner for more cores, so this change will only start 2 workers

Very strange that public repos have access to 4 cores but we only get 2 workers with `-n auto`:

```
created: 2/2 workers
2 workers [145 items]
```

PR #577, opened 2026-07-20
Preserve every row in multi-file CSV responses

CyMule:
## Summary

- concatenate each uploaded file's partition rows in request order
- serialize CSV output without a leading unnamed pandas index column
- tolerate a file that produces no elements instead of failing the request
- add regression coverage for duplicate filenames and empty-element files

## Why

CSV output for a multi-file request was built by outer-merging each file's
DataFrame into the previous one (`data.merge(resp_data, how="outer")`). An outer
merge joins on the intersection of columns, so identical rows across files
collapsed into a single row and legitimate duplicate rows were silently dropped
— most visibly when two uploads shared a filename. The merge also emitted the
row index as a leading unnamed column.

## Impact

Multi-file CSV responses now retain every row from every file (including
duplicates and repeated filenames), union differing columns, and no longer
contain the extra unnamed index column. A file that produces no elements
contributes no rows rather than failing the whole request.

## Validation

- `pytest test_general/api/test_app.py::test_output_format_csv_concatenates_multiple_files_without_index_column` — 1 passed
- `pytest test_general/api/test_app.py::test_output_format_csv_keeps_rows_when_one_file_has_no_elements` — 1 passed


paulkarayan (approved):
nits:
- the CHANGELOG/description says results "union differing columns," but neither new test exercises files with different column sets. should be a quick llm addition
- line 753 item

prepline_general/api/general.py
```diff
@@ -742,18 +742,17 @@ def join_responses(
...
-        return PlainTextResponse(data.to_csv())
+        # -- a document that produced no elements serializes to a bodiless CSV, which pandas
+        # -- cannot parse; it contributes no rows --
+        frames = [
+            pd.read_csv(io.BytesIO(response.body))  # pyright: ignore[reportUnknownMemberType]
+            for response in responses
+            if response.body.strip()
+        ]
+        if not frames:
+            return PlainTextResponse(responses[0].body)
```
paulkarayan, line 753: nit: 
do you need the response? PlainTextResponse("") seems like it would suit

paulkarayan (approved):

PR #578, opened 2026-07-20
Stop reflecting invalid API keys

CyMule:
## Summary

- return a generic `API key is invalid` detail for authentication failures
- keep the submitted credential out of both the HTTP response body and the error logs
- add regression coverage asserting the generic response and that the key is absent from captured logs

## Why

The 401 detail interpolated the submitted API key (`f"API key {api_key} is invalid"`),
and that exception is also logged by the request path — so the value landed in both the
HTTP response body and the aggregated logs. A *rejected* key is not necessarily a
non-secret: it's frequently a real credential that simply didn't match (mistyped, trailing
whitespace, wrong environment, or the wrong value pasted into the header). Treating the
header value as untrusted-but-sensitive and never reflecting or logging it avoids turning
an auth failure into a credential-disclosure event.

## Impact

The status code and authentication behavior are unchanged. Clients now receive
`API key is invalid` instead of a message echoing the supplied value, and the
credential no longer appears in logs.

## Validation

- `pytest test_general/api/test_app.py::test_general_api_returns_401` — 1 passed (asserts the generic detail and that the submitted key is absent from captured log output)


prepline_general/api/general.py
```diff
@@ -654,7 +654,7 @@ def general_partition(
         api_key = request.headers.get("unstructured-api-key")
         if api_key != api_key_env:
             raise HTTPException(
-                detail=f"API key {api_key} is invalid", status_code=status.HTTP_401_UNAUTHORIZED
+                detail="API key is invalid", status_code=status.HTTP_401_UNAUTHORIZED
```
paulkarayan, line 657: nit: you could just slice the first 4 chars if you wanted. but i dont think that's actually very beneficial. 
we know pretty well what the error is from this already.

paulkarayan (approved):
lgtm and thank you for a good find.

